PR #5, opened 2020-07-06
Implement per-call compression

peterebden:
Optimises things a bit for files that are basically incompressible (zips, jars, etc) by not wasting time trying to gzip them.


grpcutil/dial.go
```diff
@@ -94,3 +87,24 @@ func (cred tokenCredProvider) GetRequestMetadata(ctx context.Context, uri ...str
...
+}
+
+// unaryCompressionInterceptor compresses all outgoing RPCs unless SkipCompressionKey is set.
+func streamCompressionInterceptor(ctx context.Context, desc *grpc.StreamDesc, cc *grpc.ClientConn, method string, streamer grpc.Streamer, opts ...grpc.CallOption) (grpc.ClientStream, error) {
+	return streamer(ctx, desc, cc, method, compressionInterceptor(ctx, opts)...)
+}
+
+func compressionInterceptor(ctx context.Context, opts []grpc.CallOption) []grpc.CallOption {
+	if md, ok := metadata.FromIncomingContext(ctx); !ok || len(md.Get(SkipCompressionKey)) == 0 {
+		return append(opts, grpc.UseCompressor(gzip.Name))
```
Tatskaari, line 103: Does the client have to have similar smarts to know if it's compressed or not or does it just look at the message and figure it out?

peterebden: The client makes the decision in download.go (in shouldCompress) based on the filename.  Could potentially get a bit smarter with that but this seems to work well enough for now

PR #6, opened 2020-07-07
Fix process timeouts

peterebden:
They seem to be hanging due to how we use bytes.Buffer. This fixes.


mettle/worker/worker.go
```diff
@@ -549,6 +545,52 @@ func (w *worker) execute(action *pb.Action, command *pb.Command) *pb.ExecuteResp
...
+	ch := make(chan error)
+	go func() {
+		ch <- cmd.Wait()
+	}()
+	select {
+	case err := <-ch:
+		return err
+	case <-time.After(timeout):
+		log.Warning("Terminating process for %s due to timeout", w.actionDigest.Hash)
+		cmd.Process.Signal(os.Signal(syscall.SIGTERM))
```
Tatskaari, line 582: you could probably use the command context.Done() for this if you were so inclined.

peterebden: I'm not creating it with CommandContext any more; bit :scared_hamster: of that one after all this.

PR #16, opened 2020-07-15
Better timeout resilience

peterebden:
Basically put timeouts on the individual RPCs plus a rolling timeout on streaming RPCs between each message.

Also adds a concept of server liveness whereby 'dead' servers can be taken out of the pool so we don't have to keep timing them out.

### DIFF
--- a/flair/rpc/rpc.go
+++ b/flair/rpc/rpc.go
@@ -10,6 +10,7 @@ import (
 	"io"
 	"regexp"
 	"sync"
+	"time"
 
 	apb "github.com/bazelbuild/remote-apis/build/bazel/remote/asset/v1"
 	pb "github.com/bazelbuild/remote-apis/build/bazel/remote/execution/v2"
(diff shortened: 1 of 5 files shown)


flair/rpc/rpc.go
```diff
@@ -545,7 +570,33 @@ func (s *server) Delete(ctx context.Context, req *ppb.DeleteRequest) (*ppb.Delet
...
 	})
 }
+
+// withRollingTimeout implements a timeout between each step of a streamed RPC.
+// Essentially we want to fail if it takes > 5s (for example) to get a chunk, but we don't
+// want to put an upper limit on the total time the RPC can run for (it might be long
+// if it's a big file and things are moving kinda slowly).
+// It returns a new context which will be cancelled when the rolling timeout expired and
+// a callback to update its timeout.
+func withRollingTimeout(ctx context.Context, timeout time.Duration) (context.Context, func()) {
```
Tatskaari, line 586: I guess we have a general really long timeout set for gRPC in general (did you say 2 min). I suppose theoretically we could get a streaming endpoint sending us heartbeats while making no progress.

flair/trie/replication.go
```diff
@@ -53,25 +66,29 @@ func (r *Replicator) Sequential(key string, f ReplicatedFunc) error {
 // because they're all inline.
 func (r *Replicator) SequentialAck(key string, f ReplicatedAckFunc) error {
 	var e error
+	success := false
 	offset := 0
 	for i := 0; i < r.Replicas; i++ {
-		shouldContinue, err := f(r.Trie.GetOffset(key, offset))
+		shouldContinue, err := r.callAck(f, r.Trie.GetOffset(key, offset))
 		if !r.shouldRetry(err) && !shouldContinue {
```
Tatskaari, line 73: Might read better if we rename shouldRetry to isContinueable. 
```
!(r.isContinuable(err) && shouldConinue() ) {
   return err
}
```

This also leaves us the opportunity to introduce the idea of errors we can retry on the same replica.

flair/rpc/rpc.go
peterebden: I don't think these ones actually do that though; they should always be sending chunks of actual content.
I'm not too worried about one that is doing that, it's more if the server just vanishes halfway through the stream and we're getting literally nothing more - or to be able to time it out upfront if initially there's no response without limiting the whole stream in the same way.

flair/trie/replication.go
peterebden: Yeah that is a better name, agreed.

PR #21, opened 2020-07-16
Add go.mod to please servers

Tatskaari:


.plzconfig
```diff
@@ -1,3 +1,7 @@
+[please]
+location = ~/.please
+version = 15.0.0-beta9
```
peterebden, line 3: maybe `version = >=15.0.0-beta9`? then we won't get nagged when we release newer ones...

PR #22, opened 2020-07-16
Nicer cli hashes

peterebden:
Rather than passing as `--hash` and `--size`, parse them like URL fragments from the browser (which is what I often have if I'm pasting in, and is intuitive enough otherwise).

For example:
```
plz run //discern -- show -s localhost:7775 79d6c0218ddf8ab4869a69b5cc613ceb86d89ffc419637301e53b8ff950489f5/146
```
vs before.
```
plz run //discern -- show -s localhost:7775 --hash 79d6c0218ddf8ab4869a69b5cc613ceb86d89ffc419637301e53b8ff950489f5 --size 146
```

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -15,7 +15,7 @@ var log = cli.MustGetLogger()
 var actionRe = regexp.MustCompile("([0-9a-fA-F]+)/([0-9]+)/?")
 var shortActionRe = regexp.MustCompile("([0-9a-fA-F]+)")
 
-// A Action represents a combined hash / size pair written like
+// An Action represents a combined hash / size pair written like
 // ff17a4efe382e245491d6a9f1ac6bf3adce454f7e4a5559a3579c3856edf1381/122
 // This is a bit more concise than passing them with flags.
 type Action struct {
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -7,7 +7,7 @@ import (
 	"strconv"
 
 	pb "github.com/bazelbuild/remote-apis/build/bazel/remote/execution/v2"
-	"github.com/peterebden/go-cli-init"
+	"github.com/peterebden/go-cli-init/v2"
 )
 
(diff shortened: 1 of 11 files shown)


Tatskaari:
Ah I thought this before. It's annoying when copying around.

cli/cli.go
```diff
@@ -0,0 +1,69 @@
...
+	"regexp"
+	"strconv"
+
+	pb "github.com/bazelbuild/remote-apis/build/bazel/remote/execution/v2"
+	"github.com/peterebden/go-cli-init"
+)
+
+var log = cli.MustGetLogger()
+
+var actionRe = regexp.MustCompile("([0-9a-fA-F]+)/([0-9]+)/?")
```
Tatskaari, line 15: nit: Don't really need a regex here. We could just strings.split on /

cli/cli.go
```diff
@@ -0,0 +1,69 @@
...
+// An Action represents a combined hash / size pair written like
+// ff17a4efe382e245491d6a9f1ac6bf3adce454f7e4a5559a3579c3856edf1381/122
+// This is a bit more concise than passing them with flags.
+type Action struct {
+	Hash string
+	Size int
+}
+
+func (a *Action) UnmarshalFlag(in string) error {
+	matches := actionRe.FindStringSubmatch(in)
```
Tatskaari, line 27: if strings.contains(in, "/") would do it

cli/cli.go
peterebden: Yeah but the regex also handles the restricted character sets; I don't think there's a trivial way of doing that.

I'd do something else if performance was an issue but it's only CLI parsing so it's not like we're gonna need to consume thousands of these.

cli/cli.go
peterebden: yeah but then I'd have to do this anyway to get the submatches

PR #23, opened 2020-07-17
Fixed issue where purity would fail to identify broken ARs when the o…

Tatskaari:
when the output tree had missing blobs, the AR was not marked as broken as it should've been.

### DIFF
--- a/purity/gc/gc.go
+++ b/purity/gc/gc.go
@@ -451,7 +451,7 @@ func (c *collector) RemoveSpecificBlobs(hashes []string) error {
 	if c.dryRun {
 		log.Notice("Would remove %d actions:", len(hashes))
 		for _, h := range hashes {
-			log.Notice("    %s", h)
+			log.Debug("Would remove action %s", h)
 		}
 		return nil
 	} else if len(hashes) == 0 {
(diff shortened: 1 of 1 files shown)


purity/gc/gc.go
```diff
@@ -444,7 +449,10 @@ func (c *collector) shouldDelete(ar *ppb.ActionResult) bool {
 
 func (c *collector) RemoveSpecificBlobs(hashes []string) error {
 	if c.dryRun {
-		log.Notice("Would remove %d actions", len(hashes))
+		log.Notice("Would remove %d actions:", len(hashes))
+		for _, h := range hashes {
+			log.Notice("    %s", h)
```
peterebden, line 454: I think this is probably a bit too much? If it's gonna delete thousands of actions there will be a lot of this.

Maybe Info or Debug so it doesn't show up by default but you can still get them from the file?

Tatskaari: Sounds reasonable. I was tempted to print to stdout so you can pipe this to a file or back into `purity delete`. The goal being that you know exactly what you're about to delete (and don't have to calculate it again) rather than hoping the dry run and the real run delete the same stuff.

PR #28, opened 2020-07-21
Special-case the empty blob in more places

peterebden:
I thought this was already done tbh. Guess not.


elan/rpc/rpc.go
```diff
@@ -46,6 +46,9 @@ const timeout = 2 * time.Minute
 
 var log = cli.MustGetLogger()
 
+// emptyHash is the sha256 hash of the empty file.
+const emptyHash = "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
```
Tatskaari, line 50: nit: could do `var emptyHash = digest.Empty.Hash`

peterebden: Done (although mildly sad it now isn't a constant)

PR #30, opened 2020-07-22
Fix hang in FindMissingBlobs

peterebden:
Need to tick empty blobs off in the WaitGroup too.

Interestingly this implies the SDK is _not_ optimising them out on upload.


Tatskaari:
qq

PR #31, opened 2020-07-28
Add option for secondary storage that archives instead of deleting.

peterebden:
This allows an option to keep them around in case GC does the wrong thing. It transparently re-downloads to the primary on reads that can only be fulfilled from the secondary, and deletes are interpreted as moves from primary to secondary.

Some faffing around with interfaces to make the gocloud stuff work sensibly.


elan/mux/mux.go
```diff
@@ -0,0 +1,110 @@
...
+type Mux struct {
+	primary, secondary *blob.Bucket
+}
+
+func (m *Mux) Exists(ctx context.Context, key string) (bool, error) {
+	if exists, err := m.primary.Exists(ctx, key); exists && err == nil {
+		return true, nil
+	}
+	return m.secondary.Exists(ctx, key)
+}
```
Tatskaari, line 31: Could this theoretically say something doesn't exist because of a transient error with the primary? I think if the primary returns an error, we should return an error rather than checking the secondary.

peterebden: Yeah fair enough, I did wonder about this. Changed.

elan/mux/mux.go
```diff
@@ -0,0 +1,110 @@
...
+	if err != nil || offset != 0 {
+		return r, err
+	}
+	w, err := m.primary.NewWriter(ctx, key, nil)
+	if err != nil {
+		// This isn't fatal, we can still satisfy the read off the secondary.
+		log.Warning("Failed to duplicate blob to primary: %s", err)
+		return r, nil
+	}
+	return &reader{r: r, w: w}, nil
```
Tatskaari, line 49: Does this leave us open to partly written blobs? I guess that's not a problem but I thought i would check. Seems we're relying on the caller to subsequently create range-readers for all the byte-ranges of the blob in order to correctly write it back to the primary.

peterebden: I think it is OK because we always write to the end; above we check the offset and return there if it's not zero.
I don't think we have any use cases where someone writes a partial blob that isn't to the end (they might not start at the beginning if the upload was broken partway, in which case this duplication won't happen, but that shouldn't break anything).

PR #32, opened 2020-07-29
Change the 'not found' hash format to match BB format so you can actu…

Tatskaari:
…ally check if the blob exists


elan/rpc/rpc.go
```diff
@@ -323,6 +323,9 @@ func (s *server) Read(req *bs.ReadRequest, srv bs.ByteStream_ReadServer) error {
 	defer func() { <-s.limiter }()
 	r, err := s.readBlob(ctx, s.key("cas", digest), req.ReadOffset, req.ReadLimit)
 	if err != nil {
+		if gcerrors.Code(err) == gcerrors.NotFound {
+			return status.Errorf(codes.NotFound, "Blob cas/%s/%d not found", digest.Hash, digest.SizeBytes)
```
peterebden, line 327: Should maybe be a bit careful that this isn't exactly the same path that we use to store it (we have a two-digit prefix too). Dunno if that could become confusing.

PR #71, opened 2021-01-07
Just read to end of blob, don't try to calculate it

peterebden:
This can definitely do the wrong thing for compressed blobs.


Tatskaari (approved):
I think this is strictly better but I don't think I really understand how it was causing a problem before.

peterebden:
A combination of us calculating a size that's larger than the compressed blob, and not honouring the offset/limit semantics properly I think. Still more to be done here but I'm pretty sure this is better for now.

PR #97, opened 2021-03-10
Bootstrap Mettle api servers

Hamishpk:
I've removed the http endpoint and replaced it with grpc. The Serve function now makes a call to get all in flight executions before serving.


mettle/api/api.go
```diff
@@ -72,17 +69,29 @@ func serve(opts grpcutil.Opts, name, requestQueue, responseQueue, preResponseQue
 		preResponses: common.MustOpenTopic(preResponseQueue),
 		jobs:         map[string]*job{},
 	}
+
 	go srv.Receive()
+	conn, err := grpcutil.Dial(fmt.Sprintf("%s:%d", opts.Host, opts.Port), true, opts.CertFile, opts.TokenFile)
```
peterebden, line 74: Consider moving nearly all your new code into `getExecutions`; that will make this function a bit neater since it's all logically connected, and will clean up your error handling.

mettle/api/api.go
```diff
@@ -72,17 +69,29 @@ func serve(opts grpcutil.Opts, name, requestQueue, responseQueue, preResponseQue
...
+	if err != nil {
+		log.Warning("Failed to connect to bootstrap client: %s", err)
+	}
+	bootstrapClient := bpb.NewBootstrapClient(conn)
+	srv.jobs, err = getExecutions(bootstrapClient)
+	if err != nil {
+		log.Warningf("%s", err)
+	}
+	log.Notice("Updated server with %s inflight executions", len(srv.jobs))
+	conn.Close()
```
peterebden, line 84: A good Go practice is to `defer` things like this; after you've opened the connection successfully, `defer conn.Close()`, then it gets cleaned up on any exit.

proto/mettle.proto
```diff
@@ -0,0 +1,22 @@
...
+
+
+service Bootstrap {
+
+    rpc ServeExecutions(ServeExecutionsRequest) returns (ServeExecutionsResponse);
+}
+
+message Job {
+    string ID = 1;
+    bytes Current = 2;
```
peterebden, line 13: Ahahaha ok this is how you got around compiling the third-party protos. Seems legit.

FYI there is also `google.protobuf.Any` which is a _slightly_ nicer version of the same, but this is fine for what we need.

Hamishpk: haha this would solve all my problems but importing that throw a "Import "google/protobuf/any.proto" was not found or had errors." I've just tested the same thing on core and it imports fine so not sure what the issue is.

mettle/api/api.go
Hamishpk: Thanks. I'm not sure whether we should add conTLS to the grpcutl.opts? Doesn't seem like it belongs there but also adding an extra param to getExecutions seems unnecessary.

mettle/api/api.go
```diff
@@ -72,17 +70,23 @@ func serve(opts grpcutil.Opts, name, requestQueue, responseQueue, preResponseQue
 		preResponses: common.MustOpenTopic(preResponseQueue),
 		jobs:         map[string]*job{},
 	}
+
 	go srv.Receive()
+	srv.jobs, err = getExecutions(opts, true)
```
peterebden, line 75: I would probably write this as
```
if jobs, err := getExecutions(opts, true); err != nil {
   ...
} else {
    srv.jobs = jobs
}
```
It's slightly more code but safer; it would be easy to return a nil map from `getExecutions`; in general you don't necessarily expect other values to be usable when your `err` is non-nil.

mettle/api/api.go
```diff
@@ -72,17 +70,23 @@ func serve(opts grpcutil.Opts, name, requestQueue, responseQueue, preResponseQue
...
+	if err != nil {
+		log.Warningf("%s", err)
+	}
+	log.Notice("Updated server with %s inflight executions", len(srv.jobs))
```
peterebden, line 79: Also put this in the successful branch, it's a little weird to log it immediately after the warning if you _didn't_ update the executions

mettle/api/api.go
```diff
@@ -91,19 +95,59 @@ type server struct {
...
+	}
+	log.Notice("Serving %d inflight executions", len(executions))
+	return res, nil
+}
+
+// getExecutions requests a list of currently executing jobs over grpc and updates the server.
+func getExecutions(opts grpcutil.Opts, conTLS bool) (map[string]*job, error) {
+	conn, err := grpcutil.Dial(fmt.Sprintf("%s:%d", opts.Host, opts.Port), conTLS, opts.CertFile, opts.TokenFile)
+	if err != nil {
+		return map[string]*job{}, fmt.Errorf("Failed to connect to bootstrap client: %s", err)
```
peterebden, line 125: After my suggestion above, I'd `return nil, fmt.Errorf...` from these.

mettle/api/api.go
```diff
@@ -91,19 +95,59 @@ type server struct {
...
+	}
+	defer conn.Close()
+	client := bpb.NewBootstrapClient(conn)
+	log.Notice("Requesting inflight executions...")
+	req := &bpb.ServeExecutionsRequest{}
+	res, err := client.ServeExecutions(context.Background(), req)
```
peterebden, line 131: Add a timeout to this; we don't want to wait forever (e.g. on k8s we'll be talking to a statefulset, if there are no live pods in it I'm not sure if the request will fail immediately or blackhole forever, but better to be safe).
```
ctx, cancel := context.WithTimeout(context.Background(), 10 * time.Second)
defer cancel()
// pass in ctx to ServeExecutions
```

mettle/api/api.go
```diff
@@ -91,19 +95,59 @@ type server struct {
...
+		if err := proto.Unmarshal(j.Current, current); err != nil {
+			log.Warningf("unable to unmarshal s%", j.ID)
+			continue
+		}
+		job := &job{
+			Current:   current,
+			SentFirst: j.SentFirst,
+			Done:      j.Done,
+		}
+		jobs[j.ID] = job
```
peterebden, line 147: nit: just `jobs[j.ID] = &job{ ... }`, you don't use the local variable for anything

proto/mettle.proto
peterebden: This looks like https://github.com/thought-machine/please/issues/1641 (funny timing...)

As discussed, let's not worry for now. I don't think it would make a significant difference really.

mettle/api/api.go
```diff
@@ -91,19 +95,59 @@ type server struct {
...
-	e.SetIndent("", "  ")
-	e.Encode(resp)
+	res := &bpb.ServeExecutionsResponse{
+		Jobs: executions,
+	}
+	log.Notice("Serving %d inflight executions", len(executions))
+	return res, nil
+}
+
+// getExecutions requests a list of currently executing jobs over grpc and updates the server.
```
peterebden, line 121: comment is slightly inaccurate - it doesn't update the server, that's (correctly) done outside this function.

peterebden (commented):
Thanks, looks good overall, just a few small things!

mettle/api/api.go
```diff
@@ -72,17 +69,24 @@ func serve(opts grpcutil.Opts, name, requestQueue, responseQueue, preResponseQue
 		preResponses: common.MustOpenTopic(preResponseQueue),
 		jobs:         map[string]*job{},
 	}
+
 	go srv.Receive()
+	if jobs, err := getExecutions(opts, true); err != nil {
```
peterebden, line 74: why is conTLS always true? We would not want this internally for example.

peterebden (commented):
Also have a look at `grpcutil/interceptors.go`, you'll need to add this new RPC to the list in there as unauthenticated.

mettle/api/api.go
peterebden: yeah `grpcutil.Opts` is very much the server side options. There isn't an equivalent for client options (maybe there should be but not worth doing in this PR)

mettle/api/api.go
Hamishpk: Good point, this would usually be false right? I've added a TLS flag to the api options and set default: false.

peterebden (approved):
Thanks!

PR #102, opened 2021-03-23
Fix check for illegal output paths.

fische:
The current check makes builds with output path containing `..` (e.g. `..16.03.1999` or `16..03.1999`) fail while these should work


mettle/common/common.go
```diff
@@ -142,7 +142,7 @@ func checkOutputPaths(paths []string) error {
 
 // CheckPath checks that an individual input or output path doesn't contain any illegal entities.
 func CheckPath(path string) error {
-	if strings.Contains(path, "..") || strings.HasPrefix(path, "/") {
+	if strings.Contains(path, "../") || strings.HasPrefix(path, "/") {
```
peterebden, line 145: I'm not sure this is right. What if it was just `..` for example?

fische: Ah yup, I guess we can do something cleaner with the `filepath` package. I'll have a look

peterebden: Thanks! I guess we can probably simulate with a few `strings` checks if we have to...

PR #104, opened 2021-03-25
Skip updating the cache on skip cache lookup

Tatskaari:
I don't think this is quite correct but seems to improve things. At least now we get the "correct" action result back when we do `--rerun` or `--rebuild`

The spec suggests that `SkipCacheLookup` can be used to overwrite the existing action results in-case the cache is poisoned. Because our results are immutable, elan doesn't respect this. 

One way to solve this properly would be to:
1) Allow `w.client.UpdateActionResult` to actually update the action result. 
2) Have mettle check the action cache before executing an action. Mettle can then return the result from the action cache if it exists rather than executing the action. 
3) Skip that step if this flag is set, and update the action cache with our new result.

### DIFF
--- a/mettle/worker/worker.go
+++ b/mettle/worker/worker.go
@@ -551,14 +551,7 @@ func (w *worker) execute(req *pb.ExecuteRequest, action *pb.Action, command *pb.
 	log.Notice("Uploaded outputs for %s", w.actionDigest.Hash)
 	w.metadata.WorkerCompletedTimestamp = toTimestamp(time.Now())
 
-	if req.SkipCacheLookup {
-		return &pb.ExecuteResponse{
-			Status: &rpcstatus.Status{Code: int32(codes.OK)},
-			Result: ar,
-		}
-	}
-
-	ar, err = w.client.UpdateActionResult(&pb.UpdateActionResultRequest{
+	newAr, err := w.client.UpdateActionResult(&pb.UpdateActionResultRequest{
 		InstanceName: w.instanceName,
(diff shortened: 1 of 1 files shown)


mettle/worker/worker.go
```diff
@@ -550,6 +550,14 @@ func (w *worker) execute(action *pb.Action, command *pb.Command) *pb.ExecuteResp
 	uploadDurations.Observe(end.Sub(execEnd).Seconds())
 	log.Notice("Uploaded outputs for %s", w.actionDigest.Hash)
 	w.metadata.WorkerCompletedTimestamp = toTimestamp(time.Now())
+
+	if req.SkipCacheLookup {
+		return &pb.ExecuteResponse{
```
peterebden, line 555: so I don't think this is quite right, we should still try to update it in case one doesn't exist, but if `SkipCacheLookup` we should return our one rather than what the server sent back.

Having said that, that is still not quite right, because we should be able to tell the server to overwrite the action which we currently can't / don't.

Tatskaari: Okay, I've updated that. I think the correct way to go about this is as I described in the PR description. Let me know hat you think.

peterebden: Returning the existing result is deliberate to improve determinism (and I think is implicit in the API by the fact that the RPC returns an ActionResult, which suggests it doesn't have to be the one you sent over).

Think this looks good as is, but agreed we should figure out a way to allow overwriting those results in at least some cases.

peterebden (approved):

PR #105, opened 2021-03-26
Use cache priority capabilities to indicate whether we should update action results

peterebden:
Follow-on from #104 

Essentially we use the ResultsCachePolicy to communicate whether it should update or not. If SkipCacheLookup=true then we set it to a more urgent priority which Elan interprets as 'must overwrite'.

This should implement the semantics we want for these updates.


elan/rpc/rpc.go
```diff
@@ -204,6 +204,18 @@ func (s *server) GetCapabilities(ctx context.Context, req *pb.GetCapabilitiesReq
...
+					{
+						MinPriority: 0,
+						MaxPriority: 0,
+					},
+					{
+						MinPriority: -1,
+						MaxPriority: -1,
+					},
+				},
+			},
```
Tatskaari, line 218: Ah I did see this but wondered what it was about!

Tatskaari (approved):
Nice! I think this makes sense based on the description on the proto message.

elan/rpc/rpc.go
peterebden: Yeah it's a bit arbitrary and we don't actually use this response, but I think it makes sense for us to advertise it given we're now attaching some semantic meaning to it.

PR #106, opened 2021-03-26
Adding prometheus counters for a couple of errors

isobelormiston:
Specifically, the 'blob not found' and blob size mismatch errors. I don't remember seeing the size mismatch ones recently but apparently they have appeared. 

This will be in addition to metrics on the please side for 'downloading tree digest' errors.

### DIFF
--- a/elan/rpc/rpc.go
+++ b/elan/rpc/rpc.go
@@ -107,10 +107,6 @@ var knownBlobCacheMisses = prometheus.NewCounter(prometheus.CounterOpts{
 	Namespace: "elan",
 	Name:      "known_blob_cache_misses_total",
 })
-var blobNotFoundErrors = prometheus.NewCounter(prometheus.CounterOpts{
-	Namespace: "elan",
-	Name:      "blob_not_found_errors_total",
-})
 var blobSizeMismatches = prometheus.NewCounter(prometheus.CounterOpts{
 	Namespace: "elan",
 	Name:      "blob_size_mismatches_total",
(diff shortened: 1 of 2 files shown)


elan/rpc/rpc.go
```diff
@@ -767,6 +778,7 @@ func batchLabel(batched, streamed bool) string {
 // handleNotFound converts an error from a gocloud error to a gRPC one for NotFound errors.
 func handleNotFound(err error, key string) error {
 	if isNotFound(err) {
+		blobNotFoundErrors.Inc()
```
peterebden, line 781: Hmmmm, we should be careful with this one; it is not an error to return NotFound from at least GetActionResult (you could also use other endpoints in the same way, although I don't think that a well-behaved client would).

We might be able to get the same info from the `grpc_server_handled_total` metric; that has the RPC name as a label so we could exclude GetActionResult.

I think a custom metric for this would be useful in the mettle worker though; where we download the input directory (in `worker.go`), if we can identify that's failed due to missing a blob.

peterebden (commented):
Thanks!

elan/rpc/rpc.go
isobelormiston: Thanks, that does make sense. I have moved it to where we download the directory in `worker.go`, I've followed through the code path and it's the same error object so we should be able to check that the error code is NotFound at this point too.

peterebden (approved):
Thanks, LGTM!

PR #113, opened 2021-04-13
Adds optional preflight check to worker

peterebden:
Basically a small but nontrivial builtin action which can be run before a worker starts to ensure it's capable of receiving builds correctly.


mettle/main.go
```diff
@@ -162,6 +164,13 @@ func main() {
 		}
 		api.ServeForever(opts.Dual.GRPC, "", requests, responses, responses, "", false)
 	} else if cmd == "worker" {
+		if opts.Worker.Preflight {
+			log.Notice("Running preflight check")
+			if err := worker.RunOne(opts.InstanceName, "mettle-one", opts.Worker.Storage.Storage, opts.Worker.Dir, opts.Worker.Cache.Dir, opts.Worker.Sandbox, opts.Worker.AltSandbox, opts.Worker.Storage.TokenFile, opts.Worker.Cache.Prefix, opts.Worker.Cache.Part, false, opts.Worker.Storage.TLS, preflight.Digest); err != nil {
+				log.Fatalf("Preflight action run failed: %s", err)
```
Hamishpk, line 170: Would it be worth including the worker name in this error message? Might make it easier to action.

PR #121, opened 2021-05-06
Fix for WaitExecution on an already finished task

peterebden:
This can potentially lead to a hang because while we send the message saying it's done we never close the gRPC stream.


Hamishpk (approved):
LGTM!

PR #122, opened 2021-05-10
Don't terminate topics on signals

peterebden:
We still want to be able to send messages out after receiving a signal (if we're in the middle of processing a build) but don't want to receive any more.

I can't see a particularly nice way of hooking it up on the API server side but honestly not sure it's especially necessary (the [gcppubsub implementation](https://github.com/google/go-cloud/blob/2330ef456904e4ed497045a20bbccf61d43f8f5c/pubsub/gcppubsub/gcppubsub.go#L369) doesn't seem to do much with it anyway).


mettle/worker/worker.go
```diff
@@ -151,6 +148,7 @@ func runForever(instanceName, requestQueue, responseQueue, name, storage, dir, c
 	if err != nil {
 		return err
 	}
+	defer w.Shutdown()
```
Hamishpk, line 151: nit: Would it be worth expanding on the name here to make it clear what's being shut down? At first look my assumption was it shutdown the worker (which obviously didn't make sense).

mettle/worker/worker.go
```diff
@@ -122,19 +122,16 @@ func RunOne(instanceName, name, storage, dir, cacheDir, sandbox, altSandbox, tok
...
-		time.Sleep(50 * time.Millisecond) // this is dodgy obvs
-		b, _ := proto.Marshal(&pb.ExecuteRequest{
-			InstanceName: instanceName,
-			ActionDigest: digest,
-		})
-		log.Notice("Sending request to build %s...", digest.Hash)
-		if err := topic.Send(context.Background(), &pubsub.Message{Body: b}); err != nil {
-			log.Fatalf("Failed to submit job to internal queue: %s", err)
-		}
-		log.Notice("Sent request to build %s", digest.Hash)
```
Hamishpk, line 136: It might be worth keeping this log line and the one above.

Hamishpk (approved):
A couple of nits but looks good to me!

mettle/worker/worker.go
peterebden: Yeah fair enough, I was thinking of it as being kind of generic but probably better just to be explicit.

mettle/worker/worker.go
peterebden: Yeah fair, I think it is less of a deal now it's synchronous but doesn't hurt to keep them.

PR #124, opened 2021-05-14
Restructure downloads to more strongly deduplicate blobs

peterebden:
This generally feels more logical and should prevent some cases of repeated fetches.


Hamishpk (approved):
LGTM!

PR #125, opened 2021-05-21
Pushing Prometheus metrics to a gateway if one is set

isobelormiston:


mettle/worker/worker.go
```diff
@@ -513,6 +517,13 @@ func (w *worker) prepareDir(action *pb.Action, command *pb.Command) *rpcstatus.S
 		if grpcstatus.Code(err) == codes.NotFound {
 			log.Notice("Incrementing blobNotFoundErrors")
 			blobNotFoundErrors.Inc()
+			if w.promGatewayURL != "" {
+				if err := push.New(
+					w.promGatewayURL, "blob_not_found_errors_total",
+				).Collector(blobNotFoundErrors).Format(expfmt.FmtText).Push(); err != nil {
```
peterebden, line 523: Rather than pushing metrics when we increment this one individually, I think we should push continuously in the background if enabled (say every few minutes or something).
Without that, if a worker never hits this error, it will never push any of the other metrics.

mettle/worker/worker.go
```diff
@@ -915,6 +933,22 @@ func (w *worker) update(stage pb.ExecutionStage_Value, response *pb.ExecuteRespo
 	return common.PublishWithOrderingKey(ctx, w.responses, body, w.actionDigest.Hash, w.name)
 }
 
+// periodicallyPushMetrics will push this worker's metrics to the gateway every 5 minutes.
+func (w *worker) periodicallyPushMetrics() {
+	if w.promGatewayURL != "" {
+		for {
+			for metric, metricName := range metrics {
```
peterebden, line 940: This doesn't look quite right, I don't think you should have to create a new Collector for every metric each time around the loop? I used to have some code in plz that did this, it's from a while back but I don't recall it doing it quite like this 🤔

mettle/worker/worker.go
isobelormiston: Whoops, yes

mettle/worker/worker.go
```diff
@@ -915,6 +924,27 @@ func (w *worker) update(stage pb.ExecutionStage_Value, response *pb.ExecuteRespo
 	return common.PublishWithOrderingKey(ctx, w.responses, body, w.actionDigest.Hash, w.name)
 }
 
+// periodicallyPushMetrics will push this worker's metrics to the gateway every 5 minutes.
+func (w *worker) periodicallyPushMetrics() {
+	if w.promGatewayURL != "" {
+		// Set up the metrics
+		var pushers []*push.Pusher
```
peterebden, line 931: Can you not have just one Pusher? From the docs it looks like you should be able to do `.Collector(metric1).Collector(metric2)` to jam them all into the same one.

peterebden (commented):
Thanks! Didn't realise the Gatherer stuff had been deprecated, but haven't looked at this in a while!

mettle/worker/worker.go
isobelormiston: Yep I think you're right, for some reason I thought they all needed different job names. Updated now to use one pusher, thanks again :)

mettle/worker/worker.go
```diff
@@ -915,6 +924,24 @@ func (w *worker) update(stage pb.ExecutionStage_Value, response *pb.ExecuteRespo
...
+// periodicallyPushMetrics will push this worker's metrics to the gateway every 5 minutes.
+func (w *worker) periodicallyPushMetrics() {
+	if w.promGatewayURL != "" {
+		// Set up the metrics
+		pusher := push.New(w.promGatewayURL, "mettle")
+		for _, metric := range metrics {
+			pusher = pusher.Collector(metric).Format(expfmt.FmtText)
+		}
+		// Push to the gateway every 5 minutes
+		for {
```
peterebden, line 936: One more thing - would be nicer to use a `time.Ticker` here rather than sleeping each time

mettle/worker/worker.go
```diff
@@ -915,6 +924,24 @@ func (w *worker) update(stage pb.ExecutionStage_Value, response *pb.ExecuteRespo
...
+			if err := pusher.Push(); err != nil {
+				log.Warningf("Error pushing to Prometheus pushgateway: %s", err)
+			}
+			time.Sleep(5 * 60 * time.Second)
```
peterebden, line 940: 5 * time.Minute :)

peterebden (approved):
Thanks!

PR #126, opened 2021-05-25
Adds support for a Redis cache

peterebden:
This is take 2, I'm not sure if it is quite the right idea though - it might be better to tie it into Elan directly (which we may want for the GetDirectoryTree RPC)


mettle/worker/redis.go
```diff
@@ -0,0 +1,235 @@
...
+	// go r.updateActionResult(ar)
+	return r.elan.UpdateActionResult(ar)
+}
+
+func (r *redisClient) updateActionResult(ar *pb.UpdateActionResultRequest) {
+	ctx, cancel := context.WithTimeout(context.Background(), r.timeout)
+	defer cancel()
+	if err := r.redis.Set(ctx, "ar/"+ar.ActionDigest.Hash, ar.ActionResult, 0).Err(); err != nil {
+		log.Warning("Failed to set action result in Redis: %s", err)
+	}
```
Hamishpk, line 85: Why wouldn't we return the err here?

mettle/worker/redis.go
```diff
@@ -0,0 +1,235 @@
...
+func (r *redisClient) Healthcheck() error {
+	return r.elan.Healthcheck()
+}
+
+func (r *redisClient) ReadBlob(dg *pb.Digest) ([]byte, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), r.timeout)
+	defer cancel()
+	cmd := r.redis.Get(ctx, dg.Hash)
+	if err := cmd.Err(); err == redis.Nil {
+		// Blob doesn't exist in Redis
```
Hamishpk, line 59: Will we not want to know if the blob isn't in redis so we can add it in? Like a present/found flag?

peterebden: Yeah good spot - have restructured most of the logic in here to handle that more sensibly.

mettle/worker/redis.go
```diff
@@ -0,0 +1,235 @@
...
+		keys[i] = entry.Digest.Hash
+	}
+	blobs := r.readBlobs(keys, false)
+	if blobs == nil {
+		return r.elan.UploadIfMissing(entries)
+	}
+	for i, blob := range blobs {
+		if blob == nil {
+			missing = append(missing, entries[i])
+			e := entries[i]
```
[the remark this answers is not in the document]
peterebden: yup, thx!

.plzconfig
```diff
@@ -1,5 +1,5 @@
 [please]
-version = >=16.0.0-prerelease.14
+version = >=16.0.1
```
Hamishpk, line 2: haha much more civilised!

peterebden: Indeed :) There was some kind of bug in go_module which was the broken builds...

Hamishpk (approved):

PR #132, opened 2021-06-17
Do a cache check in the API server

peterebden:
We now have access to the action cache, so we can do this fairly easily. I would normally expect it to have happened already but seems worth trying again since it should be cheap.


Hamishpk (approved):
LGTM!

PR #135, opened 2021-06-27
Clean up sequential replication logic a bit

peterebden:
This doesn't really change much logically, but should report errors better (basically will show them all always). Also ups the frequency of rechecking downed servers a bit.


flair/trie/replication.go
```diff
@@ -65,21 +66,24 @@ func (r *Replicator) Sequential(key string, f ReplicatedFunc) error {
 // This facilitates the BatchReadBlobs endpoint that basically never returns an 'actual' error
 // because they're all inline.
 func (r *Replicator) SequentialAck(key string, f ReplicatedAckFunc) error {
-	var e error
+	var me *multierror.Error
```
Hamishpk, line 69: Super nit: I think merr is a better name for multierrors. It's what they use in the docs also.

Hamishpk (approved):
One nit but looks good to me. Multi errors look pretty terrible in Kibana so errorCode() is a really nice idea 8-)

flair/trie/replication.go
peterebden: Ah fair enough - guess that better matches the classic `err`!

PR #137, opened 2021-07-12
Only cache the action result if it contains all output paths

Tatskaari:


mettle/worker/worker.go
```diff
@@ -674,6 +685,38 @@ func (w *worker) execute(req *pb.ExecuteRequest, action *pb.Action, command *pb.
 	}
 }
 
+func containsAllOutputPaths(cmd *pb.Command, ar *pb.ActionResult) bool {
+	paths := make(map[string]struct{}, len(ar.OutputDirectories) + len(ar.OutputFiles))
```
peterebden, line 689: nit: you're missing the symlink fields here

peterebden (approved):

peterebden (approved):

PR #138, opened 2021-07-28
Show outputs in Discern

peterebden:
It seems useful to show this stuff as well in case a client is having issues downloading the outputs.


Hamishpk (approved):
LGTM!

PR #139, opened 2021-08-02
Allow authing downloads in Zeal

peterebden:
Basically we can apply arbitrary (non-secret) headers and also secret `Authorization` headers (both of which are needed to authenticate e.g. a Github download from a private repo).

I've tested this locally with a temporary remote_file and it's able to fetch OK.


zeal/main.go
```diff
@@ -39,7 +46,22 @@ for the Paladin skill in Diablo II since its job is to bang things down as fast
...
 func main() {
-	_, info := flags.ParseFlagsOrDie("Zeal", &opts, &opts.Logging)
-	go flags.ServeAdmin(opts.Admin, info)
-	rpc.ServeForever(opts.GRPC, opts.Storage.Storage, opts.Storage.TLS, opts.Parallelism)
+	_, info := cli.ParseFlagsOrDie("Zeal", &opts, &opts.Logging)
+	go cli.ServeAdmin(opts.Admin, info)
+
+	for domain, headers := range opts.Headers {
+		for name, header := range headers {
+			log.Notice("Applying header for %s: %s: %s", domain, name, header)
```
Hamishpk, line 54: Is there meant to be a function call or something to go with this log line?

peterebden: Nope - just noting what the config is that we loaded (since the flag structure is fairly complex it seems useful to have a record of it at startup)

peterebden: I've changed the log message so it's a little more appropriate about what it's doing (or not doing) at this point.

Hamishpk (approved):

PR #141, opened 2021-08-25
Set GOMAXPROCS automatically at startup

peterebden:
This sets it automatically based on cgroups (if we're running within one).

We did this for Please [here](https://github.com/thought-machine/please/pull/2026) after some good results internally. Also makes sense to do it for Mettle.


Hamishpk (approved):
LGTM!

PR #142, opened 2021-09-16
Fixing List() to take most recent LastAccessed time

isobelormiston:
Updating the List() function so that it uses the most recent LastAccessed time when there are two replicas of an action result.

I think this wasn’t working before because even though `existing.LastAccessed = ar.LastAccessed` is called to update the LastAccessed time, resp.ActionResults was never updated with this `existing` action result object.

### DIFF
--- a/flair/rpc/rpc.go
+++ b/flair/rpc/rpc.go
@@ -616,15 +616,20 @@ func (s *server) List(ctx context.Context, req *ppb.ListRequest) (*ppb.ListRespo
 				ar.Replicas = 1
 			}
 			if existing, present := ars[ar.Hash]; !present {
-				resp.ActionResults = append(resp.ActionResults, ar)
 				ars[ar.Hash] = ar
 			} else {
-				if existing.LastAccessed < ar.LastAccessed {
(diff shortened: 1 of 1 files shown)


flair/rpc/rpc.go
```diff
@@ -616,15 +616,20 @@ func (s *server) List(ctx context.Context, req *ppb.ListRequest) (*ppb.ListRespo
...
 			} else {
-				if existing.LastAccessed < ar.LastAccessed {
-					existing.LastAccessed = ar.LastAccessed
+				// Update ars with the replica, and use the more recent lastAccessed time.
+				ar.Replicas += existing.Replicas
+				if ar.LastAccessed < existing.LastAccessed {
+					ar.LastAccessed = existing.LastAccessed
 				}
-				existing.Replicas += ar.Replicas
+				ars[ar.Hash] = ar
```
peterebden, line 626: Hmmm so I think this was supposed to work by the action results in the map being pointers, so updating them in-place should still be reflected without us needing to re-set them in the map.

Maybe I'm missing something though?

isobelormiston:
Closing as this isn't the problem

PR #145, opened 2021-09-23
Fix data race in streamEvents.

fische:
While investigating an issue with builds stuck in executing, I noticed there's a potential data race in `streamEvents`. If I understand the code correctly, it should be harmless at the moment.

### DIFF
--- a/mettle/api/api.go
+++ b/mettle/api/api.go
@@ -280,9 +280,8 @@ func (s *server) WaitExecution(req *pb.WaitExecutionRequest, stream pb.Execution
 // streamEvents streams a series of events back to the client.
 func (s *server) streamEvents(digest *pb.Digest, ch <-chan *longrunning.Operation, stream pb.Execution_ExecuteServer) error {
 	for op := range ch {
-		opCopy := *op
-		opCopy.Name = digest.Hash
-		if err := stream.Send(&opCopy); err != nil {
+		op.Name = digest.Hash
+		if err := stream.Send(op); err != nil {
 			log.Warning("Failed to forward event for %s: %s", digest.Hash, err)
 			s.stopStream(digest, ch)
 			return err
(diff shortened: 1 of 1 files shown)


peterebden (commented):
I'm not sure I see how this was a race before? The `Send` call should be synchronous in regard to the Go-side argument passed to it.

fische:
> I'm not sure I see how this was a race before? The `Send` call should be synchronous in regard to the Go-side argument passed to it.

The issue is not with `Send` but with the assignment to `op.Name`. Since multiple routines get the same pointer to `op`, they all try to write to the same `op.Name`.

peterebden:
> The issue is not with `Send` but with the assignment to `op.Name`. Since multiple routines get the same pointer to `op`, they all try to write to the same `op.Name`.

Ah right. I am not really certain if that line is needed - but if we want to be safe it seems like we could do it once in `process` since apparently proto types are now not safe to copy (I feel like that's a new thing?).

fische:
> Ah right. I am not really certain if that line is needed - but if we want to be safe it seems like we could do it once in `process` since apparently proto types are now not safe to copy (I feel like that's a new thing?).

Yeah good point, I'm wondering if we wont have the same issue in the `process` function though. I'll see if this is possible.

mettle/api/api.go
```diff
@@ -435,7 +435,12 @@ func (s *server) process(msg *pubsub.Message) {
 					recover() // Avoid any chance of panicking from a 'send on closed channel'
 				}()
 				log.Debug("Dispatching update for %s", key)
-				ch <- op
+				ch <- &longrunning.Operation{
+					Name:     op.Name,
+					Metadata: op.Metadata,
+					Done:     op.Done,
+					Result:   op.Result,
+				}
```
fische, line 443: I dont really like this, but I dont think there's any other way to copy `op` since it contains a mutex.

peterebden (approved):

PR #146, opened 2021-09-28
Only delete job if it hasn't been overwritten

Hamishpk:
If an action runs within 5 minutes of the same action completing the job gets deleted even though the entry in the map is replaced. This PR ensures a job is deleted only when the pointer to the job is the same as it was when deletejob is called

### DIFF
--- a/mettle/api/api.go
+++ b/mettle/api/api.go
@@ -443,19 +443,21 @@ func (s *server) process(msg *pubsub.Message) {
 		}
 		if op.Done {
 			log.Info("Job %s completed by %s", key, worker)
-			go s.deleteJob(key)
+			go s.deleteJob(key, j)
 			currentRequests.Dec()
 		}
 	}
 }
 
 // deleteJob waits for a period then removes the given job from memory.
-func (s *server) deleteJob(hash string) {
+func (s *server) deleteJob(hash string, j *job) {
 	time.Sleep(retentionTime)
 	log.Debug("Removing job %s", hash)
 	s.mutex.Lock()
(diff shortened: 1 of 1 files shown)


mettle/api/api.go
```diff
@@ -443,19 +443,21 @@ func (s *server) process(msg *pubsub.Message) {
...
-func (s *server) deleteJob(hash string) {
+func (s *server) deleteJob(hash string, j *job) {
 	time.Sleep(retentionTime)
 	log.Debug("Removing job %s", hash)
 	s.mutex.Lock()
 	defer s.mutex.Unlock()
-	delete(s.jobs, hash)
+	if s.jobs[hash] == j {
+		delete(s.jobs, hash)
+	}
```
fische, line 460: Would be good to add a comment here

mettle/api/api.go
```diff
@@ -443,19 +443,22 @@ func (s *server) process(msg *pubsub.Message) {
...
 
 // deleteJob waits for a period then removes the given job from memory.
-func (s *server) deleteJob(hash string) {
+func (s *server) deleteJob(hash string, j *job) {
 	time.Sleep(retentionTime)
 	log.Debug("Removing job %s", hash)
 	s.mutex.Lock()
 	defer s.mutex.Unlock()
-	delete(s.jobs, hash)
+	// Check the action hasn't been replaced since deledJob was called
```
peterebden, line 458: `deledJob` -> `deleteJob`

peterebden (approved):

PR #151, opened 2021-12-13
Increase current requests metric when job was created by another server.

Hamishpk:
The currentRequests metric is currently sitting around -1.7M. This is due to the gauge being increased once by the replica that starts the job, then decreased buy all replicas when removing it.

This PR calls currentRequests.Inc() whenever a sever registers a job that was started by another replica.


peterebden (approved):
Nice fix, thanks!

PR #155, opened 2022-02-11
Push metrics from Elan to prom gateway

Hamishpk:
Would like to access metrics from the in office elan services to monitor network traffic between the workers and elan.

Adding a function to push metrics the the prom gateway if a URL is provided.


isobelormiston (approved):

Hamishpk:
Abandoning in favour of scraping these metrics from the bobs

PR #158, opened 2022-03-18
Add OutputFileSymlinks to list of generated outputs

Hamishpk:
Hoping this may solve the issue we are seeing where ARs are caching locally but not remotely. 

We now collect all the same outputs as [Please does](https://sourcegraph.iap.tmachine.io/github.com/thought-machine/please@a4a6124f6bbdb72ee51d959c11708fb64581b178/-/blob/src/remote/action.go?subtree=true#L382:1) before checking against expected outputs.

I've also change the set implementation to a map of bools as I feel it reads nicer.


isobelormiston (approved):

mettle/worker/worker.go
```diff
@@ -715,28 +715,32 @@ func (w *worker) execute(req *pb.ExecuteRequest, action *pb.Action, command *pb.
 }
 
 func containsAllOutputPaths(cmd *pb.Command, ar *pb.ActionResult) bool {
-	paths := make(map[string]struct{}, len(ar.OutputDirectories)+len(ar.OutputFiles)+len(ar.OutputSymlinks)+len(ar.OutputDirectorySymlinks))
+	paths := make(map[string]bool, len(ar.OutputDirectories)+len(ar.OutputFiles)+len(ar.OutputSymlinks)+len(ar.OutputDirectorySymlinks)+len(ar.OutputFileSymlinks))
```
peterebden, line 718: Don't think it really matters here but the empty struct is more efficient - it takes no space whereas the bool will need a byte (or probably more for alignment)

I also tend to prefer it to be clear that it has no meaning. Do agree the check is a bit simpler though. Guess maybe we'll get a first class set with generics!

mettle/worker/worker.go
```diff
@@ -715,28 +715,32 @@ func (w *worker) execute(req *pb.ExecuteRequest, action *pb.Action, command *pb.
...
+		paths[o.Path] = true
 	}
 
-	for _, d := range ar.OutputDirectorySymlinks {
-		paths[d.Path] = struct{}{}
+	for _, o := range ar.OutputDirectorySymlinks {
+		paths[o.Path] = true
+	}
+
+	for _, o := range ar.OutputFileSymlinks {
```
peterebden, line 737: I guess it makes sense to check all these, although IIRC the two specific symlink types are obsolete now.

peterebden (approved):
I'm not holding out much hope here since ActionResults aren't content addressible, but worth trying anyway!

mettle/worker/worker.go
Hamishpk: Yeah I'm not sure about this either. We started seeing the bug at the same time this function was added. Both local and remote checks should be the same anyway so I thought it worth adding.

PR #159, opened 2022-03-30
Add temporary logging for missing output paths

isobelormiston:
This will be useful as extra info to see why ARs aren't being cached properly.


Hamishpk:
I wonder if it might make less noise if we check the response message on the server side and add this to the log line [here](https://github.com/thought-machine/please-servers/blob/master/mettle/api/api.go#L411)? We could then remove the warning before returning the uncached response too.

Hamishpk (approved):

PR #167, opened 2022-06-09
Nack-ing the pubsub request when a worker shuts down

isobelormiston:
Following a sigterm, workers currently die after 0.5 seconds or after the current task completes. This PR introduces:
- an `immediateShutdown` flag which kills the worker immediately if set
- a 5 minute timer so the worker dies after 5 minutes even if a task is in progress
- the pubsub request is now `Nack()`'d when the worker shuts down, so that there's no 10 minute wait before it's re-tried on another worker. 

A problem with the existing shutdown is that the pubsub ack deadline is set to 10 mins when a worker picks up a task, and if it shuts down without completing the task we always wait for 10 minutes before the task is picked up by another worker.


mettle/worker/worker.go
```diff
@@ -199,12 +212,13 @@ func runForever(instanceName, requestQueue, responseQueue, name, storage, dir, c
...
-		if _, err := w.RunTask(ctx); err != nil {
-			if ctx.Err() != nil {
-				// Error came from a signal triggered above. Give a brief period to send reports then die.
-				time.Sleep(500 * time.Millisecond)
-				return fmt.Errorf("terminated by signal")
-			}
+		_, err := w.RunTask(ctx)
+		if ctx.Err() != nil {
+			// Error came from a signal triggered above. Give a brief period to send reports then die.
+			time.Sleep(500 * time.Millisecond)
```
Hamishpk, line 218: Would be cool if we can unacknowledge the message when this context is cancelled so if it's been shutdown and not completed it'll be republished straight away. Might be a little tricky here though as this function has not knowledge of pubsub messages

isobelormiston: I agree, will have a think about this

mettle/worker/worker.go
```diff
@@ -182,12 +182,25 @@ func runForever(instanceName, requestQueue, responseQueue, name, storage, dir, c
...
+		} else {
+			log.Warning("Received signal %s, shutting down when the task completes or times out...", sig)
+		}
+
+		t := time.NewTicker(5 * time.Minute)
+		defer t.Stop()
+		for {
+			select {
+			case <-t.C:
+				log.Warning("Exceeded timeout, shutting down task now")
```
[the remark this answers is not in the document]
isobelormiston: As discussed, this should now shut down immediately when there's nothing running and shut down as soon as any in-progress task has completed.

Hamishpk (approved):
LGTM

mettle/worker/worker.go
```diff
@@ -134,8 +134,8 @@ func init() {
...
-func RunForever(instanceName, requestQueue, responseQueue, name, storage, dir, cacheDir, browserURL, sandbox, altSandbox, lucidity, promGatewayURL, tokenFile, redis, redisPassword string, redisTLS bool, cachePrefix, cacheParts []string, clean, secureStorage bool, maxCacheSize, minDiskSpace int64, memoryThreshold float64, versionFile string, costs map[string]mettlecli.Currency, ackExtension time.Duration) {
-	if err := runForever(instanceName, requestQueue, responseQueue, name, storage, dir, cacheDir, browserURL, sandbox, altSandbox, lucidity, promGatewayURL, tokenFile, redis, redisPassword, redisTLS, cachePrefix, cacheParts, clean, secureStorage, maxCacheSize, minDiskSpace, memoryThreshold, versionFile, costs, ackExtension); err != nil {
+func RunForever(instanceName, requestQueue, responseQueue, name, storage, dir, cacheDir, browserURL, sandbox, altSandbox, lucidity, promGatewayURL, tokenFile, redis, redisPassword string, redisTLS bool, cachePrefix, cacheParts []string, clean, secureStorage bool, maxCacheSize, minDiskSpace int64, memoryThreshold float64, versionFile string, costs map[string]mettlecli.Currency, ackExtension time.Duration, immediateShutdown bool) {
```
peterebden, line 137: oh man these have gotten long - I really should have added some kind of options struct or something!

mettle/main.go
```diff
@@ -61,23 +61,24 @@ var opts = struct {
...
+		Sandbox           string                  `long:"sandbox" description:"Location of tool to sandbox build actions with"`
+		AltSandbox        string                  `long:"alt_sandbox" description:"Location of tool to sandbox build actions with that don't explicitly request it"`
+		Timeout           flags.Duration          `long:"timeout" hidden:"true" description:"Deprecated, has no effect."`
+		MinDiskSpace      flags.ByteSize          `long:"min_disk_space" default:"1G" description:"Don't accept builds unless at least this much disk space is available"`
+		MemoryThreshold   float64                 `long:"memory_threshold" default:"100.0" description:"Don't accept builds unless available memory is under this percentage"`
+		VersionFile       string                  `long:"version_file" description:"File containing version tag"`
+		Costs             map[string]cli.Currency `long:"cost" description:"Per-second costs to associate with each build action."`
+		ImmediateShutdown bool                    `long:"immediate_shutdown" default:"false" description:"True if the worker should shut down immediately on a sigterm."`
```
peterebden, line 77: This is a nice thought but I'm not sure it necessarily has to be a flag? Is there a case where we want the immediate shutdown behaviour?

mettle/worker/worker.go
```diff
@@ -171,23 +171,34 @@ func RunOne(instanceName, name, storage, dir, cacheDir, sandbox, altSandbox, tok
...
-		sig = <-ch
-		w.Report(false, false, false, "Received another signal %s, shutting down immediately...", sig)
-		log.Fatalf("Received another signal %s, shutting down immediately", sig)
+		if immediateShutdown || w.actionDigest == nil {
+			w.forceShutdown(fmt.Sprintf("Received shutdown signal %s, shutting down...", sig))
+		}
+		log.Warning("Received signal %s, shutting down when the task completes or times out", sig)
+		shutdownCh <- true
+
+		t := time.NewTicker(5 * time.Minute)
```
peterebden, line 192: I don't think you need a ticker here? IIUC it only happens once so this could be
```
select {
case <-time.After(5 * time.Minute):
     ....
case sig := <-ch:
    ....
}
```

mettle/worker/worker.go
```diff
@@ -199,19 +210,20 @@ func runForever(instanceName, requestQueue, responseQueue, name, storage, dir, c
...
 		runtime.GC()
 		w.Report(true, false, true, "Awaiting next task...")
-		if _, err := w.RunTask(ctx); err != nil {
-			if ctx.Err() != nil {
-				// Error came from a signal triggered above. Give a brief period to send reports then die.
-				time.Sleep(500 * time.Millisecond)
-				return fmt.Errorf("terminated by signal")
-			}
+		_, err := w.RunTask(context.Background())
+		if err != nil {
```
peterebden, line 214: nit: `if _, err := w.RunTask(context.Background()); err != nil {`
It's better to scope errors within the statement if possible.

Also assume it's deliberate to change this not to propagate the context?

peterebden (commented):
Thanks! This seems good but I'm a little unsure why it wasn't working before - it was intending to handle this kind of thing. I think the Nack() call is a good idea for sure, is that part of what's helping it?

mettle/worker/worker.go
Hamishpk: I'll make a ticket for this

mettle/main.go
Hamishpk: This is being set for lofi. We've had a number of complaints about waiting 10 minutes for mettle workers to shutdown.

Hamishpk: There could be, from my very limited systemd knowledge I feel the current setup should just terminate the worker. We have TimeoutStopSec=600 and no ExecStop so a SIGTERM _should_ be sent right away, but on occasion it counts down from 10m before doing so.

mettle/worker/worker.go
isobelormiston: Yes it is, as these changes mean it no longer cancels context to shut the worker down.

isobelormiston:
This adds the `Nack`, the 5 minute task completion timeout and the `immediateShutdown` option.

In the end the `Nack` is the most important bit I think - I originally thought that the context cancellation killed any in-flight tasks, but actually it isn't propagated to `runTask` so the tasks should already be completing before the worker shuts down.

Looking again, we might actually be able to keep the context cancellation instead of using  `shutdownCh` channel to shut the worker down once it's finished the task.

mettle/worker/worker.go
isobelormiston: I've now added the ctx back in (it's simpler than my `shutdownCh` idea after all!) so this takes in the context, as before.

mettle/main.go
isobelormiston: The 10 minute wait for tasks to complete should hopefully be fixed by the `Nack()` addition, as it'll no longer wait 10 mins for the request to be acked. This immediateShutdown option should mean that we just cancel everything straight away though - it looks to me like beforehand it would have waited for the task to finish before the worker shuts down, but perhaps there's something I've missed!

peterebden: Okay let's give it a go - I'm still suspicious that there's something weird going on with the systemd environment that's causing it but hopefully this helps!

PR #168, opened 2022-06-30
Update lucidity metrics when marking worker as unhealthy.

fische:
We've recently noticed that prometheus metrics about healthy workers reported by Lucidity were different than the one shown on the website. This is because the endpoint to fetch the list of workers changes their health as well without updating the metrics.

This PR changes that endpoint to update metrics and makes the Update endpoint check the version as well.


isobelormiston (approved):
LGTM, thanks!

PR #176, opened 2022-08-04
Make stdout & stderr uploading more robust

peterebden:
It was assuming it'd get nil for the digest if the upload fails; I don't think that is actually the case. Think it's safest just to error if we fail to upload this.

Also altered GC check to include the stdout/stderr digests when checking for any missing blobs.


Hamishpk (approved):
Might be helpful to add some logs if we error

peterebden:
> Might be helpful to add some logs if we error

Good idea, thanks.

PR #177, opened 2022-08-09
Fix concurrent map writes in purity

Hamishpk:
Purity clean is panicing due to concurrent map writes. This PR takes the collection of output digests out of markReferencedBlobs into it's own function so we can lock/unlock when necessary.

### DIFF
--- a/purity/gc/gc.go
+++ b/purity/gc/gc.go
@@ -273,29 +273,10 @@ func (c *collector) markReferencedBlobs(ar *ppb.ActionResult) error {
 		log.Debug("Found failed action result %s: exit code %d", ar.Hash, result.ExitCode)
 		c.markBroken(ar.Hash, ar.SizeBytes)
 	}
-	size := ar.SizeBytes
-	digests := []*pb.Digest{}
-	for _, d := range result.OutputDirectories {
-		sz, dgs, err := c.markTree(d)
-		if err != nil {
-			if status.Code(err) == codes.NotFound {
-				return err
-			}
-			log.Warning("Couldn't download output tree for %s, continuing anyway: %s", ar.Hash, err)
-		}
-		size += sz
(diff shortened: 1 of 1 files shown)


peterebden (approved):

purity/gc/gc.go
```diff
@@ -308,24 +289,50 @@ func (c *collector) markReferencedBlobs(ar *ppb.ActionResult) error {
 	// Mark all the inputs as well. There are some fringe cases that make things awkward
 	// and it means things look more sensible in the browser.
 	inputSize, dirs, _ := c.inputDirs(dg)
-	c.mutex.Lock()
-	defer c.mutex.Unlock()
-	c.referencedBlobs[ar.Hash] = struct{}{}
-	for _, f := range result.OutputFiles {
-		c.referencedBlobs[f.Digest.Hash] = struct{}{}
-		size += f.Digest.SizeBytes
-	}
```
fische, line 317: This part is done before `inputDirs` with your changes. Is this OK? i.e. is the order important here or not?

fische (changes requested):

purity/gc/gc.go
```diff
@@ -308,24 +289,50 @@ func (c *collector) markReferencedBlobs(ar *ppb.ActionResult) error {
...
 	for _, d := range dirs {
 		c.markDirectory(d)
 	}
```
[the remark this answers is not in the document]
Hamishpk: Good spot! I've added a lock/unlock there :)

purity/gc/gc.go
Hamishpk: I can't see that ordering is important here. It looks like it was originally done here so c.referencedBlobs was covered by the lock. Now that's been moved to it's own function it shouldn't be an issue

fische (approved):

PR #180, opened 2022-08-16
Retry storage access on internal errors

fische:
This revision adds a `retryBucket` struct that wraps the `bucket` interface in order to implement a retry logic on top of it. We go through this logic only on internal errors.

We are implementing this following an increase in the number of internal errors coming from reads on GCP buckets. The only suggested solution is to just retry when those errors occur: https://github.com/googleapis/google-cloud-go/issues/784

### DIFF
--- a/elan/rpc/retry.go
+++ b/elan/rpc/retry.go
@@ -1,0 +1,98 @@
+package rpc
+
+import (
+	"context"
+	"fmt"
+	"io"
+
+	"gocloud.dev/blob"
+	"gocloud.dev/gcerrors"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// retryBucket implements a retry logic on top of all bucket methods that can
+// return an error. Those methods are retried only on internal errors that are
+// not caused by a context cancellation.
+type retryBucket struct {
+	bucket
+	retries int
+}
+
+func (b retryBucket) NewRangeReader(ctx context.Context, key string, offset, length int64, opts *blob.ReaderOptions) (io.ReadCloser, error) {
+	var (
+		rc  io.ReadCloser
+		err error
+	)
+	err = b.retry(ctx, func() error {
+		rc, err = b.bucket.NewRangeReader(ctx, key, offset, length, opts)
+		return err
+	})
+	return rc, err
+}
+
+func (b retryBucket) NewWriter(ctx context.Context, key string, opts *blob.WriterOptions) (io.WriteCloser, error) {
+	var (
+		wc  io.WriteCloser
+		err error
+	)
+	err = b.retry(ctx, func() error {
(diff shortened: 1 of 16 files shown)


elan/rpc/retry.go
```diff
@@ -0,0 +1,98 @@
...
+		err = f()
+	}
+	if err != nil && r > 0 {
+		return fmt.Errorf("%w (retries: %d)", err, r)
+	}
+	return err
+}
+
+func (b retryBucket) isErrorRetryable(err error) bool {
+	return gcerrors.Code(err) == gcerrors.Internal || status.Code(err) == codes.Internal
```
peterebden, line 97: I'd suggest that at least ResourceExhausted and DeadlineExceeded should also be retryable. Probably also Canceled and maybe Unknown too (that's debatable but may be better to retry pointlessly than misclassify an error)

elan/rpc/retry.go
```diff
@@ -0,0 +1,98 @@
...
+// retryBucket implements a retry logic on top of all bucket methods that can
+// return an error. Those methods are retried only on internal errors that are
+// not caused by a context cancellation.
+type retryBucket struct {
+	bucket
+	retries int
+}
+
+func (b retryBucket) NewRangeReader(ctx context.Context, key string, offset, length int64, opts *blob.ReaderOptions) (io.ReadCloser, error) {
+	var (
```
peterebden, line 23: I feel like generics might provide a nicer way of sharing the implementation of these?

peterebden (commented):
Looks good but this doesn't handle the case where mettle workers go directly to storage (probably that wants a retryable wrapper around the 'clients' in this package too).

It also won't retry if you get an initial reader or writer successfully but it fails later. Do we know if that happens or if it's generally upfront that it fails?

peterebden:
You might also want to rebase on master - #181 eventually got a green build out of golangci.

elan/rpc/retry.go
fische: Yeah good point about `ResourceExhausted` and `Unknown`. Regarding `DeadlineExceeded` and `Canceled` I'll still check if this is coming from our context, in which case there's no point in retrying.

fische:
> Looks good but this doesn't handle the case where mettle workers go directly to storage (probably that wants a retryable wrapper around the 'clients' in this package too).

Do you mean when we use the `elanClient` instead of the `remoteClient` implementation? If that's what you mean, it's already using this wrapper as `New` calls `createServer`.

> It also won't retry if you get an initial reader or writer successfully but it fails later. Do we know if that happens or if it's generally upfront that it fails?

That's a very good point. There's no indication that this is only happening when initialising the reader/writers, so I'll add retries for that as well.

fische:
Abandoning this PR, as we can just use the retry logic from the GCS client, we just need to update the storage package for that.

PR #183, opened 2022-08-17
Update cloud-storage package to retry bucket access

fische:
This PR mainly updates cloud-storage and some of its dependencies, so that we can use its underlying retry logic when requests to a bucket fail.


elan/rpc/storage.go
```diff
@@ -27,6 +28,12 @@ func mustOpenStorage(url string) bucket {
 	if err != nil {
 		log.Fatalf("Failed to open storage %s: %v", url, err)
 	}
+
+	var gcsClient *storage.Client
+	if bucket.As(&gcsClient) {
+		gcsClient.SetRetry()
```
peterebden, line 34: I suspect you need to update the `go.mod` file to fix this message.

peterebden (approved):

VERSION
```diff
@@ -1 +1 @@
-9.7.11
+9.7.12
```
peterebden, line 1: could also be 9.8.0 if you want

peterebden (approved):

VERSION
fische: Yeah it definitely makes more sense to bump the minor version, mb

PR #184, opened 2022-08-18
Add version metric to blob not found

Garbett1:
Adds a version label to `blob_not_found`. If this gives us useful data, will look to add the label to the rest of the metrics.


Garbett1:
Not sure what to do about the lint there, looks like the linter doesn't pick up the return in the `for` loop?

peterebden (approved):

PR #187, opened 2022-08-18
unmarshalling execution request earilier and increase logging on graceful shutdown

goddenrich:
we believe we have an issue with nacking the action on a forced shutdown. Unfortunately logging is limited so this PR adds some helpful log lines as well as unmarshaling the pubsub message earlier (so we don't have to wait for a fetch from the elan servers before knowing what the execution request is).


mettle/worker/worker.go
```diff
@@ -462,20 +462,29 @@ func (w *worker) runTask(msg *pubsub.Message) *pb.ExecuteResponse {
...
+	if err != nil {
+		log.Error("Bad request: %s", err)
+		return &pb.ExecuteResponse{
+			Result: &pb.ActionResult{},
+			Status: status(codes.FailedPrecondition, err.Error()),
+		}
+	}
 	if req != nil {
 		w.actionDigest = req.ActionDigest
 	}
```
fische, line 475: I think you can get rid of that if block now. Since if `err` is nil, `req` should not be.

mettle/worker/worker.go
```diff
@@ -541,20 +556,26 @@ func (w *worker) extendAckDeadlineOnce(ctx context.Context, client *psraw.Subscr
...
 
 // readRequest unmarshals the original execution request.
-func (w *worker) readRequest(msg []byte) (*pb.ExecuteRequest, *pb.Action, *pb.Command, *rpcstatus.Status) {
+func readRequest(msg []byte) (*pb.ExecuteRequest, error) {
 	req := &pb.ExecuteRequest{}
+	if err := proto.Unmarshal(msg, req); err != nil {
+		return nil, fmt.Errorf("Badly serialised request: %s", err)
+	}
+	return req, nil
+}
```
fische, line 565: Do we really need a separate function for this? Could we not just call `proto.Unmarshal` in the parent function?

goddenrich: yea its not really adding any value

mettle/worker/worker.go
goddenrich: good catch thanks

fische (approved):

Hamishpk:
Might be worth adding a log line to shutdown queues as well. Just in case we are closing those without nacking

mettle/worker/worker.go
```diff
@@ -395,6 +395,7 @@ type worker struct {
 
 // ShutdownQueues shuts down the internal topic & subscription when done.
 func (w *worker) ShutdownQueues() {
+	log.Notice("Shutting down queues")
```
Hamishpk, line 398: You could include the action digest in here so we can see if it's shutting down with an in progress action

PR #188, opened 2022-08-18
wait till graceful shutdown complete

goddenrich:


mettle/worker/worker.go
```diff
@@ -209,8 +212,8 @@ func runForever(instanceName, requestQueue, responseQueue, name, storage, dir, c
 		w.Report(true, false, true, "Awaiting next task...")
 		_, err := w.RunTask(ctx)
 		if ctx.Err() != nil {
-			// Error came from a signal triggered above. Give a brief period to send reports then die.
-			time.Sleep(500 * time.Millisecond)
+			// Error came from a signal triggered above. Wait for graceful shutdown to complete.
+			wg.Wait()
```
fische, line 216: We should always wait for the goroutine to exit before returning.

mettle/worker/worker.go
```diff
@@ -180,21 +181,23 @@ func runForever(instanceName, requestQueue, responseQueue, name, storage, dir, c
...
-				w.forceShutdown("Exceeded timeout, shutting down immediately...")
-			case sig = <-ch:
-				w.forceShutdown(fmt.Sprintf("Received another signal %s, shutting down immediately...", sig))
-			}
+
+		select {
+		case <-time.After(5 * time.Minute):
+			w.forceShutdown("Exceeded timeout, shutting down immediately...")
+		case sig = <-ch:
+			w.forceShutdown(fmt.Sprintf("Received another signal %s, shutting down immediately...", sig))
```
fische, line 200: I know this has not been added in this PR, but we should Nack in the main routine, not in this one. This routine should only cancel the context, so that the main routine knows it should exit. This prevents data races and makes the code easier to understand.

Garbett1:
I think this is probably out of date/not progressing, so will close

PR #189, opened 2022-08-22
Adding grpc.WithReturnConnectionError() option to grpc.Dial

Hamishpk:
We are seeing a the below error on startup from the lofi workers. Dial fails to connect but doesn't throw an actual error so the workers neither connect to lucidity or fail. 

`Err: connection error: desc = "transport: Error while dialing dial tcp:`

This PR blocks dial until a connection is made or the context times out.


grpcutil/dial.go
```diff
@@ -15,7 +16,9 @@ import (
 // If the URL is prefixed by a protocol (grpc:// or grpcs://) that overrides the TLS flag.
 func Dial(address string, tls bool, caFile, tokenFile string) (*grpc.ClientConn, error) {
 	address, tls = parseAddress(address, tls)
-	return grpc.Dial(address, append(DialOptions(tokenFile), tlsOpt(tls, caFile))...)
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+	return grpc.DialContext(ctx, address, append(DialOptions(tokenFile), grpc.WithBlock(), tlsOpt(tls, caFile))...)
```
fische, line 21: Shouldnt we use [WithReturnConnectionError](https://pkg.go.dev/google.golang.org/grpc#WithReturnConnectionError) instead?

Hamishpk: Do we mind that WithReturnConnectionError is marked as experimental? If not happy to swap it

fische: It just means it can change or be deleted in future releases. As long as we use the same version of the gRPC library, it should not cause any issue.

fische (approved):

PR #192, opened 2022-08-26
Pushed metrics include worker name as label

peterebden:
I _think_ this is safe to do but I forget the semantics of the pushgateway and if we're allowed to add labels this way.


Hamishpk (approved):
I believe we collect please metrics through the prom gateway with the version as a label so this is probably fine!

Garbett1:
I know the Prom people have feelings on constant labels added to all metrics:
https://prometheus.io/docs/instrumenting/writing_exporters/#target-labels-not-static-scraped-labels
> If you ever find yourself wanting to apply the same label to all of your metrics, stop.

I think it's more applicable when you are scraping, rather than pushing though. I think this is fine for now, but I do wonder if we should do something like dumping the metrics to a text file and scraping that (using this: https://github.com/prometheus/node_exporter#textfile-collector )

Garbett1 (approved):

peterebden:
Yeah I agree, their suggested alternatives all seem to be about the scrape config. I am not applying this when we scrape - agreed that in those cases we wouldn't want it.

I don't think we need to dump to a file here - if we could scrape that node exporter thing we could equally scrape the worker directly - but might well be useful for plz itself which can't hang around to serve metrics.

Thanks both!

PR #193, opened 2022-09-05
Unshadow cancel func

Garbett1:
The cancel function was shadowed in the common lib. We should look at how this propagates the cancel signal in practice now.


peterebden (approved):

fische (approved):
Nice find!

PR #195, opened 2022-09-08
Return close errors if they exist

Hamishpk:
This revision returns any errors when closing the writer in uploadOne, in hope that it may improve the blob not found errors.

My theory is these are occurring due to a silent error when uploading the blobs. 

compressWriter() wraps the writer with a nopCloser if the blob doesn't require compression. In this path the writer is closed by the defer function so any errors from the writer are lost. [The write operation is not guaranteed to have succeeded until close returns with no error](https://github.com/google/go-cloud/blob/v0.26.0/blob/blob.go#L391)

If this works we should see more errors when uploading outputs, which will hopefully a little easier to debug than blob not found errors.


elan/rpc/eclient.go
```diff
@@ -108,9 +108,16 @@ func (e *elanClient) uploadOne(entry *uploadinfo.Entry) error {
 	if err != nil {
 		return err
 	}
-	defer wr.Close()
+
+	defer func() {
+		if closeErr := wr.Close(); closeErr != nil {
```
peterebden, line 113: I was thinking the other way around, like
```
if closeErr := wr.Close(); err == nil {
    err = closeErr
}
```
If `err` is already not nil then I suspect replacing it with `closeErr` isn't going to make things easier to understand.

Hamishpk: Ah of course! Thanks

elan/rpc/eclient.go
```diff
@@ -108,9 +108,16 @@ func (e *elanClient) uploadOne(entry *uploadinfo.Entry) error {
 	if err != nil {
 		return err
 	}
-	defer wr.Close()
+
+	defer func() {
+		if closeErr := wr.Close(); err == nil {
+			err = closeErr
+		}
+	}()
```
fische, line 116: Should we do this for f.Close() above as well?

Hamishpk: I think we care about the write err more but it wont hurt to add it to f.Close() also

fische (approved):

PR #196, opened 2022-09-19
Improve BatchUpdateBlobs fan-out logic in Flair

peterebden:
We were duplicating the responses for each blob; we need to deduplicate them amongst replicas (it looks like the client is interpreting it as failed if any of the replicas fail). This attempts to send back success if there is one successful replica.

Should make things more robust when one of the servers is down.


flair/rpc/rpc.go
```diff
@@ -245,12 +246,27 @@ func (s *server) BatchUpdateBlobs(ctx context.Context, req *pb.BatchUpdateBlobsR
...
-	return resp, g.Wait()
+	err := g.Wait()
+	chooseResponse := func(v []*pb.BatchUpdateBlobsResponse_Response) *pb.BatchUpdateBlobsResponse_Response {
+		for _, r := range v {
+			if r.Status.Code == int32(codes.OK) {
+				return r
+			}
+		}
+		// Not ideal but just return the first one so it has a sensible code.
+		return v[0]
```
Garbett1, line 264: I think we'd want to have some observability on when we are hitting this case. Thoughts on a log/counter here?

Garbett1 (approved):
LGTM overall, ideally we'd have some stats on when we fail all the responses, as I think that's an important signal.

flair/rpc/rpc.go
peterebden: Fair - added a log message. There aren't any metrics in Flair at all rn so I've not gone into that but we might want to think about it later (I suspect that we don't have anything that would actually collect them atm anyway).

PR #200, opened 2022-09-29
Increase logging from purity

peterebden:
List out which blobs exactly are not found when an action is missing one


peterebden:
Probably obsoleted by #201 ?

PR #201, opened 2022-09-29
Delete action results before marking blobs in purity run

Hamishpk:
This PR adds an extra step to purity Run so that we only delete blobs of action results that have actually been deleted.

New flow:
Find all blobs
Mark live action results
Delete old action results
Mark non deleted action results as live
Mark all blobs as live
Delete all over blobs

I've also changed the logic to store the cache prefix on each blob/action result. This simplifies things a little as it they don't need to be generated a second time + it means we don't need to run each hash though delete twice to cover compressed/uncompressed blobs.


proto/BUILD
```diff
@@ -2,9 +2,9 @@ sh_cmd(
     name = "generate",
     cmd = "cp $(out_location //proto/lucidity:_lucidity#protoc_go) proto/lucidity && cp $(out_location //proto/purity:_purity#protoc_go) proto/purity && cp $(out_location //proto/mettle:_mettle#protoc_go) proto/mettle",
```
Hamishpk, line 3: This and the .pb.gp files do not need review. I regenerated them to make the linter happy

purity/gc/gc.go
```diff
@@ -463,72 +463,100 @@ func (c *collector) checkDirectories(dirs []*pb.Directory) {
...
 		}
+		ch <- 1
 	}
-	for hash, size := range c.allBlobs {
+	return nil
+}
+
+func (c *collector) RemoveBlobs() error {
+	log.Notice("Determining blobs to remove...")
+	blobs := []*ppb.Blob{}
```
fische, line 512: nit: this is almost always going to close to the size of `referencedBlobs`, right? If it is, it might be a good idea to initialise the slice to `make([]*ppb.Blob, 0, len(c.referencedBlobs))`

purity/gc/gc.go
```diff
@@ -48,12 +48,14 @@ func Run(url, instanceName, tokenFile string, tls bool, minAge time.Duration, re
...
 		return err
+	} else if err := gc.RemoveActionResults(); err != nil {
+		return err
 	} else if err := gc.MarkReferencedBlobs(); err != nil {
 		return err
 	} else if err := gc.RemoveBlobs(); err != nil {
 		return err
 	} else if err := gc.RemoveBrokenBlobs(); err != nil {
-		return err
+		log.Notice("Failed to remove broken blobs: %v", err)
```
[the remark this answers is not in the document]
Hamishpk: The other functions are best effort/log errors rather than return them. RemoveBrokenBlobs is used by Clean also and needs to return an error. I put a log line here so replicate is always called regardless of the outcome of RemoveBrokenBlobs

purity/gc/gc.go
```diff
@@ -463,72 +463,100 @@ func (c *collector) checkDirectories(dirs []*pb.Directory) {
...
+	var wg sync.WaitGroup
+	wg.Add(c.parallelism + 1)
+	step := len(c.actionResults) / c.parallelism
+	for i := 0; i < (c.parallelism + 1); i++ {
+		go func(blobs []*ppb.Blob) {
+			ctx, cancel := context.WithTimeout(context.Background(), 1*time.Hour)
+			defer cancel()
+			_, err := c.gcclient.Delete(ctx, &ppb.DeleteRequest{
+				Blobs: blobs,
+			})
```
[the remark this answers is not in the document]
Hamishpk: Good spot, thanks!

purity/gc/gc.go
Hamishpk: c.referencedBlobs don't get deleted. I have set the capacity to c.allBlobs - c.referencedBlobs.

purity/gc/gc.go
```diff
@@ -275,7 +275,7 @@ func (c *collector) markReferencedBlobs(ar *ppb.ActionResult) error {
 	}
 	outputSize, digests, err := c.outputs(result)
 	if err != nil {
-		log.Warning("Couldn't download output tree for %s, continuing anyway: %s", ar.Hash, err)
+		log.Debug("Couldn't download output tree for %s, continuing anyway: %s", ar.Hash, err)
```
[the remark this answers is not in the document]
Hamishpk: The reason I have changed this to debug is because it makes the output unreadable in the terminal. It mainly gets ran on a bob though so this doesn't really matter so would be happy to change it back if you think that's better?

elan/rpc/gc.go
```diff
@@ -72,20 +73,16 @@ func (s *server) list(ctx context.Context, prefix, prefix2 string) ([]*ppb.Actio
...
 	log.Notice("Delete request for %d action results, %d blobs", len(req.ActionResults), len(req.Blobs))
-	// Note that we don't differentiate between compressed and uncompressed blobs in the request
-	// (the fact they're stored differently is an implementation detail). This is a _little_
-	// inefficient since we have to check all of them twice, but c'est la vie.
 	var g multierror.Group
-	g.Go(func() error { return s.deleteAll(ctx, "ac", req.ActionResults, req.Hard) })
-	g.Go(func() error { return s.deleteAll(ctx, "cas", req.Blobs, req.Hard) })
-	g.Go(func() error { return s.deleteAll(ctx, "zstd_cas", req.Blobs, req.Hard) })
+	g.Go(func() error { return s.deleteAll(ctx, req.ActionResults, req.Hard) })
+	g.Go(func() error { return s.deleteAll(ctx, req.Blobs, req.Hard) })
```
isobelormiston, line 78: Perhaps the other way of doing this would be to give DeleteRequest  both a `Blobs` and a `CompressedBlobs` field to distinguish them, but maybe that makes things more complicated?

Hamishpk: That would definitely make things quicker but you are right it would make things a little complicated. I've only ran this in dry_run so don't know how long the delete part actually takes. Maybe a good idea would be to land this, check the times when QAing it, then if time to complete is an issue I can raise a separate revision to speed things up?

purity/gc/gc.go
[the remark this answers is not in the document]
fische: `ReplicateBlobs` wont be called if there's an error though :sweat_smile:

purity/gc/gc.go
fische: Ah yeah, good point. Although I think you forgot to push that change

purity/gc/gc.go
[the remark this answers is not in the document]
Hamishpk: Haha oops. I've set it to just return the error

ChangeLog
```diff
@@ -1,3 +1,11 @@
+Version 9.9.4
+-------------
+    * Restructure Purity so that it deletes action results before marking
+	their blobs as live.
+	* Add cache prefix field to purity action results proto
+	* Add cache prefix field to purity blobs proto
+	* Elan List() returns blobs with cache preflix field populated 
+
```
Birkalo, line 8: nit: This should be version `10.0.0` and the bullet points could be aligned better

purity/gc/gc.go
```diff
@@ -463,72 +463,112 @@ func (c *collector) checkDirectories(dirs []*pb.Directory) {
...
-					Blobs:         blobs[key],
-					ActionResults: ars[key],
-				})
-				if err != nil {
-					me = multierror.Append(me, err)
-				}
-				ch <- 1
+	var wg sync.WaitGroup
+	wg.Add(c.parallelism + 1)
+	step := numBlobs / c.parallelism
```
Birkalo, line 547: What happens here if `numBlobs` isn't a multiple of `c.parallesism` ?

fische (approved):
LGTM! Thanks Hamish

purity/gc/gc.go
Hamishpk: The remainder is caught by the extra loop. For example if numBlobs = 10 and parallelism = 3. The step size would be 10 / 3 = 3 then we loop through parallelism + 1 (in this case 4) times which covers the remainder. Does that make sense?

PR #206, opened 2022-10-17
Fix bugs + no longer return not found errors from Elan gc

Hamishpk:
As well as fix a few silly bugs this PR removes blob not found errors from elans gc.

Not 100% sure on what's happening here but often a delete returns a not found error even though the blob has actually been deleted. Not a huge deal but is causing action results that aren't there to be marked as live which makes things messy.


VERSION
```diff
@@ -1 +1 @@
-10.0.1
+10.1.1
```
fische, line 1: We jump straight to this version? Should it not be 10.1.0?

elan/rpc/gc.go
```diff
@@ -85,15 +85,13 @@ func (s *server) deleteAll(ctx context.Context, blobs []*ppb.Blob, hard bool) er
...
-				me = multierror.Append(me, err)
+				me = multierror.Append(me, fmt.Errorf("Error deleting blob: %w", err))
 			}
 			if s.knownBlobCache != nil {
 				s.knownBlobCache.Del(key)
 			}
 		} else if err != nil {
-			me = multierror.Append(me, err)
-		} else {
-			me = multierror.Append(me, fmt.Errorf("Blob not found: %s", key))
```
fische, line 96: Why are we getting rid of this? I think this still makes sense. I'm happy if we change it to a `log.Error` instead though.

purity/gc/gc.go
```diff
@@ -492,24 +493,28 @@ func (c *collector) RemoveActionResults() error {
 	wg.Add(c.parallelism + 1)
 	step := numArs / c.parallelism
 	for i := 0; i < (c.parallelism + 1); i++ {
+		i := i
```
Birkalo, line 496: This is interesting, why do we have to assign i to i. Is it a concurrency thing?

purity/gc/gc.go
```diff
@@ -239,6 +239,7 @@ func (c *collector) MarkReferencedBlobs() error {
 	wg.Add(c.parallelism + 1)
 	step := len(c.actionResults) / c.parallelism
 	for i := 0; i < (c.parallelism + 1); i++ {
+		i := i
```
fische, line 242: Is this useful? `i` is not used in the goroutine, so this should not be required

purity/gc/gc.go
```diff
@@ -483,7 +484,7 @@ func (c *collector) RemoveActionResults() error {
 		return nil
 	}
 	log.Notice("Deleting %d action results, total size %s", numArs, humanize.Bytes(uint64(totalSize)))
-	ch := newProgressBar("Deleting blobs", 256)
+	ch := newProgressBar("Deleting action results", 256)
```
fische, line 487: Should we not use `len(ars)` instead of 256? Or at least `c.parallelism`?

purity/gc/gc.go
fische, line 496: Same comment as above, is this necessary?

purity/gc/gc.go
```diff
@@ -543,8 +548,10 @@ func (c *collector) RemoveBlobs() error {
 		time.Sleep(10 * time.Millisecond)
 	}()
 	var wg sync.WaitGroup
-	wg.Add(numBlobs)
+	wg.Add(len(blobs))
 	for k, v := range blobs {
+		k := k
+		v := v
```
fische, line 554: Same

purity/gc/gc.go
Garbett1: It's a for loop thing in go. https://github.com/golang/go/discussions/56010 This covers what they want to change about the semantics!

Garbett1: Although, I think `i` isn't used, so there's no need I think?

Hamishpk: That's correct. The nightly job occasionally fails early with no warning and I assumed it may have been an issue here. Probably due to not locking when writing to a map below though

elan/rpc/gc.go
Hamishpk: I have logged this as a warning and not returned the error. The other option here is to return custom errors and handle them on  client side but that's annoying with multierrors.

ChangeLog
```diff
@@ -1,3 +1,9 @@
+Version 10.1.1
```
fische, line 1: I think you forgot to change the version here as well :D

purity/gc/gc.go
```diff
@@ -543,7 +546,7 @@ func (c *collector) RemoveBlobs() error {
 		time.Sleep(10 * time.Millisecond)
 	}()
 	var wg sync.WaitGroup
-	wg.Add(numBlobs)
```
fische, line 546: At that point, I think you can get rid of `numBlobs`, as it's not used anywhere.. I think?

purity/gc/gc.go
```diff
@@ -496,17 +496,20 @@ func (c *collector) RemoveActionResults() error {
 			ctx, cancel := context.WithTimeout(context.Background(), 1*time.Hour)
 			defer cancel()
 			for _, ar := range ars {
-				log.Debug("Removing action result %s", ar.Hash)
```
fische, line 499: Why are we getting rid of this debug log?

elan/rpc/gc.go
fische: I think logging it makes sense, thanks! :+1:

purity/gc/gc.go
Hamishpk: We log above saying we have scheduled it for deletion so decided to remove this to try and reduce lines. the logfile produced by this is huge so thought it would be best to drop this one

purity/gc/gc.go
Hamishpk: We use it in the notice saying how many blobs we are going to delete. I think it's worth keeping

PR #207, opened 2022-10-20
Add very blunt timeout to deal with possibilities of unreliable networks

Garbett1:
While the internal REX servers are playing up, add some blunt-force timeouts on the RPC endpoints.

We need to investigate why the underlying SDK wasn't using it's default timeouts, but the current symptoms are clearly an infinite wait.

2 should be sufficient for interactions with Elan, but I must admit, I don't know the usual time bounds.


elan/rpc/rclient.go
```diff
@@ -33,39 +33,53 @@ func (r *remoteClient) Healthcheck() error {
...
+	ctx, cnx := context.WithTimeout(context.Background(), time.Minute*2)
+	defer cnx()
+	return r.c.UpdateActionResult(ctx, req)
 }
 
 func (r *remoteClient) UploadIfMissing(entries []*uploadinfo.Entry) error {
-	_, _, err := r.c.UploadIfMissing(context.Background(), entries...)
+	ctx, cnx := context.WithTimeout(context.Background(), time.Minute*2)
+	defer cnx()
+	_, _, err := r.c.UploadIfMissing(ctx, entries...)
```
peterebden, line 57: I would be a bit worried about this one. `UploadIfMissing` can potentially make a bunch of internal RPCs and while I'd generally expect us to upload outputs in under 2 minutes total, I can imagine some things (aws-sdk-go...) being slow.

Garbett1: Cool, will increase the two timeouts you had concerns about, and also whack in the metrics so we can visibly see when it's timing out.

elan/rpc/rclient.go
```diff
@@ -33,39 +33,53 @@ func (r *remoteClient) Healthcheck() error {
...
 func (r *remoteClient) ReadToFile(dg digest.Digest, filename string, compressed bool) error {
+	ctx, cnx := context.WithTimeout(context.Background(), time.Minute*2)
+	defer cnx()
 	if !compressed {
-		_, err := r.c.ReadBlobToFileUncompressed(context.Background(), dg, filename)
+		_, err := r.c.ReadBlobToFileUncompressed(ctx, dg, filename)
 		return err
 	}
-	_, err := r.c.ReadBlobToFile(context.Background(), dg, filename)
+	_, err := r.c.ReadBlobToFile(ctx, dg, filename)
```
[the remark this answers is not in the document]
Garbett1: Made this 5m as well.

fische (approved):
LGTM, thanks Xander!

Hamishpk (approved):

PR #208, opened 2022-10-24
Mark action results for purity clean

Hamishpk:
After further investigation errors seen from bucket.Exists() seem to occur when the blob doesn't exist. I have changed these to a log line so that purity only marks action results as live if bucket.Delete fails.

I've also added markActionResults as it's needed for purity clean.


ChangeLog
```diff
@@ -1,6 +1,9 @@
 Version 10.2.0
+-------------
     * Add timeouts to Elan's remote client
     * Add metrics to Elan's remote client
+	* Only return delete errors in Elan's gc
+	* Fix purity clean
```
fische, line 6: I think you have to make a new version for the release pipeline to be triggered.

Hamishpk: It looks like the release pipeline didn't produce any assets anyway so I was going to delete it before landing this

fische: But still, are you sure it's gonna get triggered when merging this? I thought it was getting triggered if the `VERSION` is changed

Garbett1: Looks like the release _attempted_ to trigger, but failed: https://github.com/thought-machine/please-servers/actions/runs/3311956028/jobs/5468074324#step:6:30

Hamishpk: The release process is pretty flaky. I'm sure I've seen that error before!

I have bumped the version in this PR

fische (approved):

PR #209, opened 2022-10-26
Fix metric cardinality

Garbett1:
Fix the mistake in the metrics, bump the version.


Hamishpk:
Will need to update the change log also so the release gets triggered

Hamishpk (approved):

PR #210, opened 2022-11-09
Add a label with worker's name to the errors counter

Liu-ko:
A new label should help to identify misbehaving workers.


Garbett1:
If we want to cut a release straight away, we should add changes to the `VERSION` and `ChangeLog` files (similar to: https://github.com/thought-machine/please-servers/commit/308eb10c5906ab9eb914595287bdbb8b91cec763 )

Otherwise, LGTM!

Garbett1 (approved):

PR #211, opened 2022-11-22
Use workers name as a job name while pushing metrics to a prom gateway

Liu-ko:
Adding a hostname for all the metrics


Garbett1:
Looks good to me! Thanks!

PR #213, opened 2022-12-16
Adding debug logs to elan and upping flair logs to errors

Hamishpk:


elan/rpc/gc.go
```diff
@@ -83,9 +83,9 @@ func (s *server) deleteAll(ctx context.Context, blobs []*ppb.Blob, hard bool) er
 	var me *multierror.Error
 	for _, blob := range blobs {
 		key := blob.CachePrefix + blob.Hash
-		log.Notice("Deleting blob %s", key)
 		if exists, err := s.bucket.Exists(ctx, key); exists {
 			if err := s.bucket.Delete(ctx, key, hard); err != nil {
+				log.Error("Error deleting blob: %v/home/hpitkeathly/.local/state/nvim/swap//%home%hpitkeathly%please-servers%elan%rpc%rpc.go.swp", err)
```
Birkalo, line 88: this looks like a copy&paste typo, but also might be something important, i have no idea

flair/trie/replication.go
```diff
@@ -84,9 +84,9 @@ func (r *Replicator) SequentialAck(key string, f ReplicatedAckFunc) error {
 		}
 		merr = multierror.Append(merr, err)
 		if i < r.Replicas-1 {
-			log.Debug("Error reading from replica for %s: %s. Will retry on next replica.", key, err)
+			log.Error("Error reading from replica for %s: %s. Will retry on next replica.", key, err)
```
Birkalo, line 87: nit: line 87, 89, 131 could have the `err` verb as  `%w`

elan/rpc/gc.go
Hamishpk: :aha: You're right that is a c&p error. That's the location of a previous vim swp file I needed to remove before I could open a new one.

flair/trie/replication.go
Hamishpk: Good spot. I have changed these to %v. Githubs UI isn't great here but you can see the changes if you hit Files changed.

Birkalo (approved):

PR #218, opened 2023-01-26
Add time to complete, total successful actions and nacked message metrics

Hamishpk:
timeToComplete and totalNackedMessages have been added to monitor the effects (if any) of moving over to spot instances.

totalSuccessfulActions is just to make our success rate calculations simpler.

I've also demoted some log lines to debug in worker.go


mettle/api/api.go
```diff
@@ -58,10 +58,29 @@ var totalFailedActions = prometheus.NewCounter(prometheus.CounterOpts{
 	Name:      "failed_actions_total",
 })
 
+var totalSuccessfulActions = prometheus.NewCounter(prometheus.CounterOpts{
+	Namespace: "mettle",
+	Name:      "Successful_actions_total",
```
fische, line 63: nit: to keep with the naming standards, it should only be lower case

isobelormiston (approved):

fische (approved):

PR #222, opened 2023-02-03
Use a tighter timeout when reading from redis.

fische:
Since reading from redis is in the critical path, we should use a smaller timeouts to avoid having all workers stuck when redis is struggling with the load.


Garbett1 (approved):
Seems good!

Hamishpk (approved):

PR #223, opened 2023-02-07
Allow using read replicas with redis client

fische:
Based on GCP memorystore docs, it looks like they use their own replication tool, so we can't use any of the implementation that cover read replicas in go-redis package.


Hamishpk (changes requested):
LGTM

Hamishpk (approved):
LGTM

PR #227, opened 2023-03-18
Fix load balancing logic

Garbett1:
Moves the LB logic to the Replicator, rather than the Trie.

Load balancing on the executor is currently disabled, but that shouldn't make much difference.

Added tests this time to ensure it all works this time :)


Garbett1:
Really not sure what's going on with golang-ci here!

flair/trie/replication.go
```diff
@@ -19,6 +20,8 @@ var log = logging.MustGetLogger()
 
 var serverDead = status.Errorf(codes.DeadlineExceeded, "Server is down")
 
+var rnd = rand.New(rand.NewSource(time.Now().UnixNano()))
```
peterebden, line 23: The `Rand` type isn't safe for concurrent use. This would be better as a member of the `Replicator` struct but I still don't think that's enough (you could get multiple calls going through the same replicator simultaneously)

peterebden (commented):
Right, sorry, should have clicked in the last review that Trie was not the place for this.

flair/trie/replication.go
Garbett1: Cool, I’ve just made the intN function aliased in the package and override in tests, which should do the trick!

peterebden (approved):
Thanks!

PR #228, opened 2023-03-18
Build fixes

peterebden:
- Updates Go plugin & Go version, and actually uses toolchain (because we can't use the system Go 1.20)
 - Adds a `go_system_toolchain` for CI builds on Alpine.
 - Fix redis compile for presumably more recent gcc versions (I think we may as well turn off `-Werror` since we aren't writing this code ourselves)
 - Fix bb_browser for linux_arm64
 - Update `golangci-lint` to fix whatever was going wrong with it.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -14,6 +14,9 @@ issues:
     - appendAssign
     - halp  # Make misspell be quiet about this.
     - exitAfterDefer # Potentially useful but not in any cases it fires right now.
+    - unused-parameter # Quite a few of these, some legit, many are fulfilling interfaces
+    - redefines-builtin # Not really a big issue
+    - empty-block # Only came up once and was a false positive. This should be easily handled by review.
   exclude-rules:
     - path: _test\.go
       linters:
@@ -22,15 +25,12 @@ issues:
 linters:
   disable-all: true
   enable:
-    - deadcode
     - gosimple
     - govet
     - ineffassign
     - staticcheck
-    - structcheck
     - typecheck
     - unused
-    - varcheck
     - asciicheck
     - bodyclose
     - dogsled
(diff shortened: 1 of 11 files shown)


peterebden:
Think this needs a new Docker image, but I don't have an appropriate machine handy right now. Hopefully will come back to this soon.

peterebden:
Ugh right. I think this is because of mixing the glibc-based Go toolchain with the Alpine image we're building in. However Go 1.20 _needs_ that to get the stdlib compiled. What a pain.

Garbett1:
Does a multistage docker file work here? Glib based for the go tool chain, grab the output and pivot into a musl one?

Garbett1:
Ah, is this Go itself expecting glib based tool chains to be available at runtime. That’s nasty.

peterebden:
Yeah the Go binaries that we download from `golang.org` are linked against glibc. They just don't run on Alpine. But for 1.20+ we have to compile the stdlib which that does for us.
I think https://github.com/please-build/go-rules/pull/85 is a reasonable fix here. Trying that out on this PR.

.golangci.yml
```diff
@@ -14,6 +14,9 @@ issues:
     - appendAssign
     - halp  # Make misspell be quiet about this.
     - exitAfterDefer # Potentially useful but not in any cases it fires right now.
+    - unused-parameter # Quite a few of these, some legit, many are fulfilling interfaces
```
peterebden, line 17: We _might_ want to have a think about this; it can be a useful check, but it fires on a lot of cases that are fulfilling interfaces etc and changing `func write(force bool) error` to `func write(_ bool) error` is a step backwards in readability.

Garbett1: Tbh, I come across these in IJ and it’s cleverer about realising that it’s due to an interface. So personally happy to disable

Garbett1:
Builds are green!

Garbett1 (approved):

PR #231, opened 2023-04-20
Add force-check on the CAS

Garbett1:
Due to mistakes, we have destroyed blobs in the CAS, but not the corresponding AR.

This is somewhat irritating to identify due to the synthetic actions still existing, so we make a change to Zeal to double check the CAS if the opt is enabled.


zeal/rpc/rpc.go
```diff
@@ -99,14 +107,29 @@ func (s *server) FetchBlob(ctx context.Context, req *pb.FetchBlobRequest) (*pb.F
...
+				return &pb.FetchBlobResponse{
+					Status:     &rpcstatus.Status{},
+					BlobDigest: ar.OutputFiles[0].Digest,
+				}, nil
+			} else {
+				//  Missed the CAS, note it down, and move on to usual times.
+				casMissing.Inc()
+			}
+		} else {
+			log.Info("Retrieved %s from action cache (as %s/%d). Did not check CAS", req.Uris, dg.Hash, dg.Size)
```
Garbett1, line 127: I'm not particularly happy with this spaghetti, but my brain is currently unable to process more complex thoughts.

zeal/rpc/rpc.go
```diff
@@ -99,14 +107,29 @@ func (s *server) FetchBlob(ctx context.Context, req *pb.FetchBlobRequest) (*pb.F
 	if ar, err := s.storageClient.CheckActionCache(ctx, dg.ToProto()); err != nil {
 		log.Error("Failed to check action cache: %s", err)
 	} else if ar != nil {
-		if len(ar.OutputFiles) == 1 {
-			log.Info("Retrieved %s from action cache (as %s/%d)", req.Uris, dg.Hash, dg.Size)
+		if len(ar.OutputFiles) != 1 {
```
Garbett1, line 110: This is buggy anyway actually, just thought about this more.

Garbett1: Fixed with the most recent change

zeal/rpc/rpc.go
```diff
@@ -99,8 +107,27 @@ func (s *server) FetchBlob(ctx context.Context, req *pb.FetchBlobRequest) (*pb.F
 	if ar, err := s.storageClient.CheckActionCache(ctx, dg.ToProto()); err != nil {
 		log.Error("Failed to check action cache: %s", err)
 	} else if ar != nil {
+		if s.forceCasCheck {
+			if resp, err := s.storageClient.FindMissingBlobs(ctx, &rpb.FindMissingBlobsRequest{
+				InstanceName: s.storageClient.InstanceName,
+				BlobDigests:  []*rpb.Digest{dg.ToProto()},
```
peterebden, line 113: I don't think this works? The digest is the digest of the message, which we're using as an arbitrary key into the AC. I don't think we expect it to exist in the CAS.

Garbett1: Updated in the latest, we can assume we have exactly 1 output file, so just grab it's digest.

fische (approved):

Garbett1:
Will add the release files to trigger a release on merge as well.

PR #232, opened 2023-04-20
Flip the boolean logic

Garbett1:
Missed this when we pushed last time. 

I tested this locally against an action that was messed up and confirmed it got pushed to Elan.


fische (approved):
:facepalm:

PR #234, opened 2023-04-25
Updates mettle to use new plz plugins etc

peterebden:
sh, proto and go_proto. Deals with some of the proto upgrades etc too


Garbett1 (approved):
LGTM!

PR #235, opened 2023-04-26
Update remote-apis-sdks

peterebden:
This unforks remote-apis so we're fully in line with their protocol on compression. Updates remote-apis-sdks (still with a bunch of custom changes because they don't support this yet, although I've sent most of it back upstream now).

Have tested locally by rebuilding mettle against the local instance, which shook out a few small bugs.

I've bumped a major version (beta) because this won't roll out in one go; we need to roll out Elan and Flair, then the Mettle workers to maximise intelligibility of the Capabilities messages.


rexclient/client.go
```diff
@@ -25,35 +24,24 @@ const CompressionThreshold = 1024
...
+		&client.TreeSymlinkOpts{Preserved: true},
+		client.CompressedBytestreamThreshold(CompressionThreshold),
+		client.UploadCompressionClassifier(client.DetectArchiveUploads),
+		client.UsePackName(PackName),
+	)
 	if err != nil {
 		log.Error("Error initialising remote execution client: %s", err)
 		return nil, err
 	}
-
```
Garbett1, line 38: So this all goes away with the new compression arguments I guess?

elan/rpc/rpc.go
```diff
@@ -571,42 +575,52 @@ func (s *server) readBlob(ctx context.Context, key string, offset, length int64)
...
-	buf := make([]byte, 0, digest.SizeBytes)
-	if compressed {
-		return s.compressor.EncodeAll(b, buf), nil
-	}
-	b, err = s.decompressor.DecodeAll(b, buf)
-	if err != nil {
-		return nil, err
-	}
-	blobsServed.WithLabelValues(batchLabel(batched, false), compressorLabel(compressed), compressorLabel(!compressed)).Inc()
+	blobsServed.WithLabelValues(batchLabel(batched, false), compressorLabel(compressed), compressorLabel(compressed)).Inc()
```
Garbett1, line 618: `compressorLabel(compressed),compressorLabel(!compressed)`

elan/rpc/eclient.go
```diff
@@ -56,11 +56,13 @@ func (e *elanClient) UpdateActionResult(req *pb.UpdateActionResultRequest) (*pb.
 	return e.s.UpdateActionResult(ctx, req)
 }
 
-func (e *elanClient) UploadIfMissing(entries []*uploadinfo.Entry) error {
+func (e *elanClient) UploadIfMissing(entries []*uploadinfo.Entry, compressors []pb.Compressor_Value) error {
```
Garbett1, line 59: Does changing this signature mean that all the Please clients will fail for a short period of time if we roll Mettle first?

peterebden: Not this particular case, no. This interface is only relevant within the process.

The upload/download blob stuff should all be compatible here, the main bit that's not is the `Capabilities` stuff (which I think we can work around by ordering the rollout; we could probably avoid it by doing it in several more steps but I don't think it's worth it)

Garbett1 (approved):
Good to go once the metric label is updated.

elan/rpc/rpc.go
peterebden: Hmmmm on some thought I've set the second one to not applicable. They're "compressor used" and "compressor preferred", in this new world there is no preferred compressor when serving batched blobs.

PR #239, opened 2023-05-18
Update SDK 

peterebden:
This is now using the upstreamed version of most code; packs are the only thing we still have changed from master on the fork.
The whole protocol should now be mutually intelligible, hence tagging this as v11


rexclient/compression.go
```diff
@@ -0,0 +1,39 @@
...
+	"net/http"
+	"strings"
+
+	"github.com/bazelbuild/remote-apis-sdks/go/pkg/uploadinfo"
+)
+
+// DetectArchiveUploads is an UploadCompressionPredicate that attempts to detect whether
+// uploads are already archives, either based on their initial few bytes or known file extensions.
+// It is a best-effort classifier that may not recognise every possible type.
+func DetectArchiveUploads(ue *uploadinfo.Entry) bool {
```
peterebden, line 14: They (probably reasonably) didn't want this upstream in the SDK, but the API lets us determine compression from this repo now.

Garbett1 (approved):
The actual moving parts are small, so the change LGTM!

PR #241, opened 2023-06-05
Implement path lookup according to REAPI v2.3

peterebden:
This changed upstream - it used to be either an absolute path on the server or relative within the working dir. They have subsequently (sensibly) changed it so paths without separators are looked up (which makes sense because the client doesn't know where `bash` is on the server).

Also populated a new field & reporting the new API version


mettle/worker/worker.go
```diff
@@ -1145,3 +1147,29 @@ func resultsCachePriority(skipCacheLookup bool) int32 {
...
+
+// commandPath returns the path that should be used for this command.
+// Per the spec, if it contains a path separator we don't alter it; if not we look it up,
+// honouring the given env var if present.
+func commandPath(command *pb.Command) string {
+	arg := command.Arguments[0]
+	if strings.ContainsRune(arg, filepath.Separator) {
+		return arg
+	}
+	for _, env := range command.EnvironmentVariables {
```
peterebden, line 1159: We have this in plz too, it's a little tedious. Might put a request in upstream to add something to the stdlib for this, if I can think of a good name for the function.

Garbett1: Seems like a reasonable impl.

Garbett1 (approved):

Garbett1:
LGTM. Anything else interesting from bumping to 2.3?

peterebden:
> LGTM. Anything else interesting from bumping to 2.3?

I don't think so. The two things in this PR are the only ones mentioned in the proto - sadly I have forgotten all the other discussions that went on in the meantime. (I think it'd help if there were tagged releases, but c'est la vie)

PR #244, opened 2023-06-13
targets updated by puku

goddenrich:


elan/rpc/BUILD
```diff
@@ -48,10 +48,9 @@ go_test(
     deps = [
         ":rpc",
         "///third_party/go/github.com_klauspost_compress//zstd",
-        "///third_party/go/github.com_stretchr_testify//assert",
-        "///third_party/go/github.com_stretchr_testify//require",
         "///third_party/go/google.golang.org_genproto_googleapis_bytestream//:bytestream",
         "//grpcutil",
+        "//third_party/go:testify",
```
Garbett1, line 53: Why do we need to refer to the installed version if the tests were working without this?

Garbett1:
I think this is not needed anymore.

PR #246, opened 2023-06-18
Fix bug in bytestream offsets

peterebden:
#242 seems to break building things in localremote with errors about `incorrect WriteOffset (was 3145728, should be 3104718)`

The issue first appears in klauspost/compress@1.14.2. Haven't bisected it further yet and I don't fully understand what is going wrong; I have some suspicion that this is just triggering a bug on our side due to a change in their behaviour (there seem to be some cases where the read buffer we're given changes size partway into a stream).

Anyway, am downgrading it to the previous working minor version for now so things work again. I'd like to understand more about what's going on here still though.


.plzconfig.localremote
```diff
@@ -6,3 +6,4 @@ NumExecutors = 10
 TokenFile = grpcutil/token.txt
 Secure = false
 DisplayUrl = http://localhost:7779
+Instance = mettle
```
peterebden, line 9: This is a minor one I found while noodling about; without it the client doesn't generate very nice URLs. It doesn't matter much otherwise since we don't do a lot with the instance name.

Garbett1 (approved):

peterebden:
Okay have updated this with the real fix & upgraded compress again - it isn't their fault :)

v1.14.2 introduces a behaviour that sometimes provides different sized read buffers - I don't fully grok the code about why they do that but I'm sure it makes sense for them. It turns out that we were never counting properly on our side - when we check against TotalSize, we might have some left in our buffer if it wasn't enough to fulfil the complete read. We have to account for that too.
Previously this wasn't happening because our internal buffer was always an exact multiple of the supplied read buffer, so it was always zero when we needed to read another chunk.

Garbett1 (approved):
Approving on the second pass to look at the TotalSize changes.

PR #247, opened 2023-06-26
Improve error handling in worker

peterebden:
This tries to be a bit more specific about error classification, specifically identifying context timeout errors as DeadlineExceeded so callers will know that they should retry.


Garbett1 (approved):
LGTM.

PR #261, opened 2023-10-18
Add configurable pubsub package so we can increase the number of publishers on topics

Hamishpk:
The number of handlers used by a topic is not configurable using the gcppubsub package. We are often seeing context deadline exceeded errors when publishing to the response queue at times when throughput is high. This error is client side and little to no deadline exceeded errors are seen on the gcp side.

This revision adds a configurable pubsub package so we configure both the max handlers and the batch size. We will now have to pass the topic url with the prefix configureable- if we want to increase the max handlers.


Hamishpk:
Not ready for review

mettle/configurablepubsub/configurablepubsub.go
```diff
@@ -0,0 +1,87 @@
...
+	"regexp"
+	"strconv"
+	"strings"
+
+	"gocloud.dev/gcp"
+	"gocloud.dev/pubsub"
+	"gocloud.dev/pubsub/gcppubsub"
+)
+
+const Scheme = "configurable-gcppubsub"
```
fische, line 18: That's a very long name for a scheme 😅

mettle/configurablepubsub/configurablepubsub.go
```diff
@@ -0,0 +1,87 @@
...
+	if err != nil {
+		return nil, err
+	}
+	topicPath := path.Join(u.Host, u.Path)
+	if topicPathRE.MatchString(topicPath) {
+		return gcppubsub.OpenTopicByPath(pc, topicPath, &opts)
+	}
+	// Shortened form?
+	topicName := strings.TrimPrefix(u.Path, "/")
+	return gcppubsub.OpenTopic(pc, gcp.ProjectID(u.Host), topicName, &opts), nil
```
fische, line 78: What about using `gcppubsub.URLOpener` here? You initialise it with the connection and topic options and then just call OpenTopicURL with the URL. That way we can still benefit from their implementation and their extra params.

mettle/api/api.go
```diff
@@ -91,36 +109,52 @@ func init() {
...
+		return nil, nil, fmt.Errorf("too few pollers specified: %d", queueOpts.NumPollers)
 	}
 	srv := &server{
 		name:         name,
-		requests:     common.MustOpenTopic(requestQueue),
-		responses:    common.MustOpenSubscription(responseQueue, responseBatchSize),
-		preResponses: common.MustOpenTopic(preResponseQueue),
+		requests:     common.MustOpenTopic(queueOpts.RequestQueue, queueOpts.TopicBatchSize, queueOpts.NumPublishers),
+		responses:    common.MustOpenSubscription(responseSubscriptionName, queueOpts.SubscriptionBatchSize),
+		preResponses: common.MustOpenTopic(queueOpts.PreResponseQueue, queueOpts.TopicBatchSize, queueOpts.NumPublishers),
```
fische, line 153: I know so far we've passed those extra options as separate flags, but what about setting them directly in the URL that we give to mettle? That way we don't have to do some weird manipulation of the URL in the code.

mettle/configurablepubsub/configurablepubsub.go
Hamishpk: haha I shorterned it to tm-gcppubsub. I wouldn't be against changing the package name to this also but the name is pretty descriptive.

peterebden:
Are we truly sure this is the issue? It seems strange that we're really hitting the limits of publish rate - how many messages are we pushing to pub/sub?

Hamishpk:
> Are we truly sure this is the issue? It seems strange that we're really hitting the limits of publish rate - how many messages are we pushing to pub/sub?

I just realised as we publish with an ordering key we want a batch size of 1. Given this we can increase the number of handlers within the server application code so I'm going to change this!

Hamishpk:
Got myself in a right rebase mess here so abandoning

PR #262, opened 2023-10-25
Group pubsub opts for api server and add metrics related to failing to

Hamishpk:
The revision groups the api server pubsub opts and adds some metrics around failure to publish messages


fische (approved):
Nice, thanks!

PR #268, opened 2023-10-29
Feat: Add rate limit to Redis client

Garbett1:
We had some issues with slow downs on workers pulling slowly from Redis. 

This PR:
a. Lowers the timeout to 1 second (which is probably reasonable for a Redis command?)
b. Adds a rate limiter in the Redis client that skips the Redis lookups if more than 10 errors have happened in the last 10 seconds.

We only add the rate limit on the read operations, because they are in the critical path.
We want to attempt to keep populating the Redis cache so we can recover to a useful state after the Redis issues.

Not adding configuration just yet, but will make some follow up PRs to add metrics on this.


mettle/worker/redis.go
```diff
@@ -91,23 +94,33 @@ func (r *redisClient) Healthcheck() error {
...
 	blob, err := r.elan.ReadBlob(dg)
 	if err != nil {
 		return nil, err
 	}
 	if dg.SizeBytes < r.maxSize {
-		ctx, cancel = context.WithTimeout(context.Background(), r.timeout)
-		defer cancel()
-		if cmd := r.redis.Set(ctx, dg.Hash, blob, 0); cmd.Val() != "OK" {
-			log.Warning("Failed to set blob in Redis: %s", cmd.Err())
+		if r.limiter.Allow() {
```
Garbett1, line 116: Not sure whether this needs to be in the rate limit given that I've moved this to be in a go routine anyway. Feels that we want to skip _all_ Redis or not in one single movement.

I suppose this might be clearer with a top level `if allow` that delegates to the Elan client and skips everything?

peterebden: It's probably good not to spin off a bunch of goroutines if we think Redis is down anyway?

rexclient/BUILD
```diff
@@ -23,6 +23,6 @@ go_test(
         ":rexclient",
         "///third_party/go/github.com_bazelbuild_remote-apis-sdks//go/pkg/digest",
         "///third_party/go/github.com_bazelbuild_remote-apis//build/bazel/remote/execution/v2",
-        "///third_party/go/github.com_stretchr_testify//assert",
```
peterebden, line 26: I think these are better not changed... the build failures seem like the subpackage builds okay but the whole thing doesn't

peterebden (approved):

mettle/worker/redis.go
```diff
@@ -91,23 +94,33 @@ func (r *redisClient) Healthcheck() error {
...
+			go func() {
+				ctx, cancel = context.WithTimeout(context.Background(), r.timeout)
+				defer cancel()
+				if cmd := r.redis.Set(ctx, dg.Hash, blob, 0); cmd.Val() != "OK" {
```
peterebden, line 120: One thing to be mildly cautious of here is that `dg` is now used after this function returns. I doubt this is likely to be a problem anywhere, but just mentioning.

rexclient/BUILD
Garbett1: I've fixed the `go.mod` so these work as expected now.

mettle/worker/redis.go
Garbett1: I've ended up removing the limiter on writes, so we keep attmping to upload to Redis, even if the reads are failing. 

The timeout is short enough at 1 second that I think this is _probably_ fine from a memory perspective.

mettle/worker/redis.go
Garbett1: I'll close over the `dg.Hash` just to remove any doubt!

PR #269, opened 2023-11-10
Start receiving from queue only once we got all inflight executions.

fische:
This should fix a potential data race when we initialise the internal job cache. We used to start receiving from the queue first and then fetch the list of inflight executions to fill the internal job cache. The first uses a mutex to write to the cache but not the second. There are 2 solutions, either we keep the same order of doing things and we make sure to lock the mutex when filling the cache with inflight executions, or we just start receiving after initialising the cache. The latter seems safer to me, which is why I went for that solution.


ChangeLog
```diff
@@ -1,3 +1,8 @@
+Version 11.6.1
```
Hamishpk, line 1: Should we hold off creating a new release until the retention/expiry is done?

Hamishpk (approved):

ChangeLog
fische: Yeah I was actually thinking the same while raising the quick fix for the retention/expiry

Garbett1: I think it's fine to make a bunch of smaller releases here, we just might skip past internally.

peterebden (approved):

PR #270, opened 2023-11-10
Randomly offset retention and expiration time.

fische:
Randomly offset retentionTime and expiryTime for each job so the goroutines don't all wake up at the same time. This is a bit of a hack, a proper fix will be raised soon.


ChangeLog
```diff
@@ -1,3 +1,9 @@
+Version 11.6.1
```
Garbett1, line 1: This is going to conflict with https://github.com/thought-machine/please-servers/pull/269 when we land, but we should see that soon.

fische: Yeah I know. As said on that PR, I'll land it later in case it breaks something

Garbett1 (approved):

PR #272, opened 2023-11-10
Don't use Allow, because that consumes tokens

Garbett1:
The previous code was consuming tokens when it meant to only check them. This meant that we effectively skipped all Redis operations.


mettle/worker/redis.go
```diff
@@ -95,7 +95,7 @@ func (r *redisClient) ReadBlob(dg *pb.Digest) ([]byte, error) {
 	ctx, cancel := context.WithTimeout(context.Background(), r.timeout)
 	defer cancel()
 
-	if r.limiter.Allow() {
+	if r.limiter.Tokens() < 1.0 {
```
peterebden, line 98: >= 1.0, no?

(the other looks correct)

Garbett1: Ah, yes. Mistakenly did both at the same time.

fische (approved):

PR #274, opened 2024-01-06
Add GCP Cloud profiler support

Garbett1:
Add optional support for sending profiling information to GCP Cloud Profiler.


.github/workflows/codeql-analysis.yml
```diff
@@ -18,7 +18,7 @@ jobs:
   analyze:
     name: Analyze
     # TODO(jpoole): set this back to ubuntu-latest once https://github.com/actions/virtual-environments/issues/1816 lands
```
peterebden, line 20: you can remove this line now

cli/cli.go
```diff
@@ -29,24 +31,41 @@ type LoggingOpts struct {
 }
 
 // AdminOpts is a re-export of the admin type so servers don't need to import it directly.
-type AdminOpts = admin.Opts
+type AdminOpts struct {
+	admin.Opts
```
peterebden, line 35: I dunno what the linter is moaning about here. Should be fine.

cli/cli.go
```diff
@@ -29,24 +31,41 @@ type LoggingOpts struct {
...
+	EnableGcpProfiling bool `long:"gcp_profiling" description:"Enable pushing profiles to GCP Cloud profiling." env:"ADMIN_GCP_PROFILING"`
+}
 
 // ParseFlagsOrDie parses incoming flags and sets up logging etc.
 func ParseFlagsOrDie(name string, opts interface{}, loggingOpts *LoggingOpts) (string, logging.LogLevelInfo) {
 	cmd := flags.ParseFlagsOrDie(name, opts)
 	info := logging.MustInitStructuredLogging(loggingOpts.Verbosity, loggingOpts.FileVerbosity, loggingOpts.LogFile, loggingOpts.Structured)
 	if _, err := maxprocs.Set(maxprocs.Logger(log.Notice), maxprocs.Min(1)); err != nil {
-		log.Error("Failed to set GOMAXPROCS: %s", err)
+		log.Errorf("Failed to set GOMAXPROCS: %s", err)
```
peterebden, line 44: I don't think it matters which of these we call, but fine either way

third_party/proto/resourceusage/BUILD
```diff
@@ -1,6 +1,9 @@
 # This is a vendorised copy of Buildbarn's resource usage proto for display in the browser.
 proto_library(
     name = "resourceusage",
-    srcs = ["resourceusage.proto"],
+    srcs = [
+        "resourceusage.proto",
+        "resourceusage.pb.go",
```
peterebden, line 6: This doesn't seem right. The proto_library emits this file

peterebden (commented):
Interesting!

I dunno what the linter is so annoyed about... it seems confused

third_party/proto/resourceusage/BUILD
Garbett1: So, this is `puku fmt` that is doing this. I had reverted that change in a previous commit, and not sure _why_ Puku feels the need to do that.

.github/workflows/codeql-analysis.yml
Garbett1: Done

cli/cli.go
Garbett1: Yeah, going to attempt to explicitly import as `admin`

peterebden (approved):
Nice!

PR #277, opened 2024-02-09
Set lowercase names for the profiler to work.

Garbett1:
GCP Cloud profiler gave me the following error:

```
service name \"Elan\" does not match regular expression ^[a-z0-9]([-a-z0-9_.]{0,253}[a-z0-9])?$
```

This lowercases all those service names, and puts a backstop in there just in case.


peterebden (dismissed):
capital letters are deprecated

fische (approved):

PR #281, opened 2024-02-20
Periodically delete jobs in Mettle api server

Hamishpk:
At times of high load the current method of scheduling actions for deletion using individual goroutines is causing connection/context deadline exceeded errors. 

This change hopes to improve this by managing the deletion of jobs through one routine. Every 10 minutes the server will loop through all the jobs and delete it if:

-  The job is complete and over the retention time (5 minutes)
-  The job is incomplete, has no listeners and within the expiry time (60 minutes)
-  The job is incomplete and outside of two times the expiry time (120 minutes)

I remember we had some nasty bugs a few years ago around this so I've added a bunch of tests for shouldDeleteJobs. I think this should cover all our cases but am slightly worried an empty job (done==false, LastUpdate == beginning of time) would be deleted. We never set an empty job so shouldn't be an issue but could be convinced to add a special case for this)


mettle/api/api.go
```diff
@@ -554,55 +553,32 @@ func (s *server) process(msg *pubsub.Message) {
...
-func (s *server) expireJob(hash string) {
-	time.Sleep(expiryTime + time.Duration(rand.Int63n(int64(expiryTime))))
-	if s.maybeExpireJob(hash, false) {
-		return
+func (s *server) periodicallyDeleteJobs() {
+	for range s.deleteJobsTicker.C {
+		for digest, job := range s.jobs {
+			if shouldDeleteJob(job) {
+				s.mutex.Lock()
+				defer s.mutex.Unlock()
```
peterebden, line 565: You shouldn't defer this here. The defer won't be run until the function exits so you'll deadlock this pretty quickly.
Either pull the inner part out to its own function or maybe just manually unlock when done (since the code inside it is very simple)

I think this is what the linter is trying to tell you about although I'm not sure why it doesn't comment  here

Hamishpk: Ahh of course! This was blindly pulled out of the old delete job, have updated to unlock once we have deleted from the map. Thanks

Hamishpk:
After an in-person chat with @fische it was decided to go for @peterebden suggestion of locking the whole map. I have added a metric to track the time of this and will refactor if it proves to be troublesome

mettle/api/api.go
```diff
@@ -554,55 +560,34 @@ func (s *server) process(msg *pubsub.Message) {
...
-	}
-}
-
-// expireJob expires an action that hasn't progressed.
-func (s *server) expireJob(hash string) {
-	time.Sleep(expiryTime + time.Duration(rand.Int63n(int64(expiryTime))))
-	if s.maybeExpireJob(hash, false) {
-		return
+func (s *server) periodicallyDeleteJobs() {
+	startTime := time.Now()
```
fische, line 568: This should be within the loop. The Observe as well

fische (changes requested):

fische (approved):

PR #283, opened 2024-03-03
Update job deletion logic + propagate update times

Garbett1:
The previous deletion logic would always check that the streams were zero-ed before deleting the job: https://github.com/thought-machine/please-servers/commit/0b744c9f805d8d17e70bc12fbc042a37d2c1ecdf#diff-030e4836d83d50a740c86eb47a2780e7a81f170cc5902279ff28ad5122081d37L593

This diff updates the logic to match the previous logic, and updates the `LastUpdated` time in a few places so it's a more accurate measurement.


peterebden (approved):
Nice find!

Hamishpk (commented):
Would you mind swapping 573 and 574? Slight bug in the metrics where it starts recording  before it locks. It's on my to do list but haven't found time :/

Hamishpk (approved):
Nice! Thanks Xander

mettle/api/api.go
```diff
@@ -372,6 +372,10 @@ func (s *server) WaitExecution(req *pb.WaitExecutionRequest, stream pb.Execution
 func (s *server) streamEvents(digest *pb.Digest, ch <-chan *longrunning.Operation, stream pb.Execution_ExecuteServer) error {
 	for op := range ch {
 		op.Name = digest.Hash
+		s.mutex.Lock()
+		j := s.jobs[digest.Hash]
+		j.LastUpdate = time.Now()
+		s.mutex.Unlock()
```
fische, line 378: Should we really update LastUpdate here? This is just sending an event to the client, not updating the actual job that we store in memory

fische:
The stream check was only on the expiration, not on the deletion routine. Not sure if we're expecting any remaining stream once the action has been completed, but it's probably harmless, even safer, just in case.

mettle/api/api.go
Garbett1: I think that makes sense right? If I have an operation that is sending progress events, for each one of them I want to make sure that the jobs liveness property is being incremented. 

If I have a job that takes 10 minutes to run for example, I don't want the `LastUpdate` to be set only at the start when the job is enqueued IMO?

Might be that we should make what `LastUpdate` actually represents clearer.

fische: > If I have a job that takes 10 minutes to run for example, I don't want the LastUpdate to be set only at the start when the job is enqueued IMO?

Why not? To me, this is why we have all those conditions in `shouldDeleteJob`:
```
	if !j.Done && len(j.Streams) == 0 && timeSinceLastUpdate > expiryTime {
		return true
	}
	if !j.Done && timeSinceLastUpdate > 2*expiryTime {
		return true
	}
```
ie if the job is still in progress, we keep it for the expiration time (1h) and while it still has some streams. Otherwise we keep it for 2h.

Garbett1: I'm specifically thinking about this condition:
```
timeSinceLastUpdate := time.Since(j.LastUpdate)
if j.Done && len(j.Streams) == 0 && timeSinceLastUpdate > retentionTime {
                // return true to delete
		return true
}
```

If my job takes 10 minutes to run, and we don't update with progress events, once that job is done we immediately remove it (even if a client might attempt to reconnect)

My understanding is that we are retaining the jobs to smooth that out.

Liu-ko (approved):

PR #284, opened 2024-03-03
Refactor gauge strategy

Garbett1:
Given that the job requests are quite a hot function, and we care only about snapshots of the requests, this refactors it to use a GuageFunc to get the length of the jobs map when it's scraped. 

The registering code needs to close over the `server`, which is the only slighy ugliness of this approach, especially needing to wrap with `sync.Once` otherwise the tests panic.

### DIFF
--- a/mettle/api/api.go
+++ b/mettle/api/api.go
@@ -173,6 +173,8 @@ func serve(opts grpcutil.Opts, name string, queueOpts PubSubOpts, apiURL string,
 			Name:      "requests_current",
 		},
 			func() float64 {
+				srv.mutex.Lock()
+				defer srv.mutex.Unlock()
 				return float64(len(srv.jobs))
 			},
 		))


mettle/api/api.go
```diff
@@ -168,6 +169,19 @@ func serve(opts grpcutil.Opts, name string, queueOpts PubSubOpts, apiURL string,
...
 	}
+
+	// _Technically_ this won't happen more than once in normal running, as we'd only run 1 server, but it does happen in tests.
+	register.Do(func() {
+		prometheus.MustRegister(prometheus.NewGaugeFunc(prometheus.GaugeOpts{
+			Namespace: "mettle",
+			Name:      "requests_current",
+		},
+			func() float64 {
+				return float64(len(srv.jobs))
```
peterebden, line 176: this could introduce a concurrent map read / modification - if the main logic is updating it at the same time as Prometheus tries to scrape it

Garbett1: D'oh. Yeah, brain did not engage on that one.

Hamishpk (approved):

fische (approved):

PR #285, opened 2024-03-03
Make platform details optional

Garbett1:
The platform check is surprisingly expensive, mainly because it requires a full on CAS read. 

For homogeneus fleets (such as the main TM fleet), this check is not particularly useful, we never really expect the platform to make a difference. The worker code itself doesn't even check if the platform is viable.

This change introduces a short curcuit if the platform properties aren't set, and just skips the CAS read for the platform.

### DIFF
--- a/mettle/api/api.go
+++ b/mettle/api/api.go
@@ -167,6 +167,7 @@ func serve(opts grpcutil.Opts, name string, queueOpts PubSubOpts, apiURL string,
 		client:           client,
 		numPollers:       queueOpts.NumPollers,
 		deleteJobsTicker: time.NewTicker(10 * time.Minute),
+		actuallyValidate: len(allowedPlatform) > 0,
 	}
 	log.Notice("Allowed platform values:")
 	for k, v := range allowedPlatform {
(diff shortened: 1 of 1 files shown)


peterebden (dismissed):

mettle/api/api.go
```diff
@@ -202,6 +203,7 @@ type server struct {
 	mutex            sync.Mutex
 	numPollers       int
 	deleteJobsTicker *time.Ticker
+	actuallyValidate bool
```
Hamishpk, line 206: Is it necessary to add an extra field on the server struct here? Could we just check the length of s.platform below?

fische: +1, also it almost feels like the check should even be part of `validatePlatform`.

Garbett1: I refactored in this way mainly because I was initially passing through a flag to ignore, rather than the approach I took in the end.

Personally, I prefer to read it this way, but not 100% wedded to it, can just in line the check.

Garbett1: The other thing is that we expect `Execute` (and therefore validatePlatform) to be called a whole bunch of times, but the platform config itself never actually changes, and I think the boolean on the struct kinda indicates that more clearly.

I do agree that it should be pushed down into `validatePlatform` actually, I think that keeps the code clearer.

Hamishpk: This makes sense to me. Thanks Xander

Hamishpk (approved):

PR #287, opened 2024-03-05
Reduce verbosity of worker logs

isobelormiston:
The existing logging for a successful action looks like this:
“Received task..”
“Executing action..” 
“Prepared directory for..” 
“Completed execution for..” 
“Uploaded outputs for..” 
“Uploaded action result for..”

This reduces it to just “Executing action..”


Hamishpk:
Should we change the other happy path notice logs to debug and ensure all error logs are up to date?

For example we don't need to know a worker uploaded all outputs but we do need to know if it failed to upload outputs.

Hamishpk (approved):

PR #289, opened 2024-03-07
Ensure done jobs get deleted periodically 

Hamishpk:
Done jobs aren't being deleted so they must have streams hanging around on the jobs. I'm removing the logic that updates the last update time when streaming events and made a change to shouldDelete jobs.

The new logic:

A job without listeners will get deleted if it is done and time since last update > retention time or if it's not done and time since last update > expriy time.

A job with listeners will get delete if it is done and time sice last update is > resumption time.

Any job that hasn't had an update in 2x expiry will be deleted.

I've also ensured j.Done is set to false when we create a new execution and added warnings around deleting jobs with listeners.


Hamishpk:
Done jobs aren't being deleted so they must have streams hanging around on the jobs. I'm removing the logic that updates the last update time when streaming events and made a change to shouldDelete jobs.

The new logic:

A job without listeners will get deleted if it is done and time since last update > retention time or if it's not done and time since last update > expriy time.

A job with listeners will get delete if it is done and time sice last update is > resumption time.

Any job that hasn't had an update in 2x expiry will be deleted.

I've also ensured j.Done is set to false when we create a new execution and added warnings around deleting jobs with listeners.

fische (approved):

PR #290, opened 2024-03-11
Only clear J.current if we reuse an old job

Hamishpk:
Previous behaviour sets j.current to nil after we set it it include some metadata a few lines above it.

I've also added some debug logs to see if this incomplete response error is happening whilst we are cleaning jobs


fische (dismissed):
Dont forget to bump the version and update the Changelog

fische (approved):

PR #292, opened 2024-03-22
Replace goroutine with unlimited buffered channel

peterebden:
As discussed, I think we have potential ordering issues by dispatching things in separate goroutines. That was necessary to avoid blocking when sending to a channel if the receiver is slow.

This removes the goroutine and replaces it with a self-buffering channel-like thing which should maintain send order.

Also did a bit of housekeeping (because I wanted to use Go 1.22 range-over-int, which needed several other updates)


peterebden:
Don't the workers send a periodic update once a minute or so?

peterebden:
We could also get rid of that, or just have a buffer size of like 100 and deal with it if they fill up

fische:
> Don't the workers send a periodic update once a minute or so?

Just had a quick look at the worker code and it looks like it only publish when [Executing](https://github.com/thought-machine/please-servers/blob/master/mettle/worker/worker.go#L606) and when [Completed](https://github.com/thought-machine/please-servers/blob/master/mettle/worker/worker.go#L438).

> We could also get rid of that, or just have a buffer size of like 100 and deal with it if they fill up

Yeah we could even do that

PR #293, opened 2024-03-25
Remove goroutine

peterebden:
Simpler version of #292 ; remove the goroutine and rely on our channels being longer than the number of responses we get (which we think should always be the case)
Should save a bit of CPU too since we're not spinning up so many routines.

Removed the channel-closing stuff so we'll never get a 'send on closed channel', we just bail out of the loop when we get one that's marked Done and let the GC clean it up later.


fische (approved):
Thanks, LGTM! We'll have to keep an eye on memory usage, to make sure those channels do get GC-ed.

fische (approved):

PR #294, opened 2024-03-25
Update to Go 1.22

peterebden:
...and related stories (some other things need bumping too)

Also some code cleanup, making better use of `slices` and `maps` now we have those.


Garbett1 (approved):
LGTM

PR #302, opened 2024-04-20
WIP: Teach Purity about instance names 

Garbett1:
Purity abuses the instance name for the REX Api to make Elan have special cases. Additionally, it doesn't pass through instance names through out.

Still need to re-generate the protos, and ran into some "fun".

### DIFF
--- a/purity/gc/gc.go
+++ b/purity/gc/gc.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/hex"
 	"fmt"
+	"google.golang.org/grpc/metadata"
 	"path"
 	"sort"
 	"sync"
@@ -124,6 +125,8 @@ type Action struct {
 	InputSize, OutputSize int
 }
 
+const name = "purity-gc"
+
 type collector struct {
 	client            *client.Client
 	gcclient          ppb.GCClient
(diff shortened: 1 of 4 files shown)


purity/gc/gc.go
```diff
@@ -187,6 +190,7 @@ func (c *collector) LoadAllBlobs() error {
 			for j := 0; j < 16; j++ {
 				ctx, cancel := context.WithTimeout(context.Background(), 30*time.Minute)
 				defer cancel()
+				ctx = metadata.NewOutgoingContext(ctx, metadata.Pairs("mettle-caller-name", name))
```
peterebden, line 193: I'm not normally that person, but maybe we should have a constant for `mettle-caller-name`...

Or possibly a top-level variable for this `Pairs` instance might be easier?

peterebden (approved):

purity/gc/gc.go
Garbett1: Was considering moving it into the common package and perhaps making it a gRPC interceptor.

Not sure what conventions gRPC has for naming metadata as well, almost went for mettle.tmachine.io/name, but that feels very Java/k8s.

peterebden: > Keys are case insensitive and consist of ASCII letters, digits, and special characters -, _, . and must not start with grpc- (which is reserved for gRPC itself). Binary-valued keys end in -bin while ASCII-valued keys do not.

So you can't have the / at least. Lower kebab case seems normal from what I've seen.

Garbett1:
I think this needed more work to actually get functional, because the REX SDK was problematic to update.

PR #303, opened 2024-04-22
Read from redis only if blob size is below limit.

fische:
@Hamishpk realised we read from redis even for blobs whose size is above the maxSize, which should therefore not be found there. This will hopefully limit the number of calls and help with the load on redis.


Hamishpk (approved):
LGTM

PR #306, opened 2024-04-29
Make BatchDownload honour maxSize when uploading to redis.

fische:
We're still seeing huge increase in MSET and SET to redis since #303 was landed and deployed. I think it is because BatchDownload does not check the blob size when uploading to redis. This PR should fix this.


Hamishpk (approved):
LGTM

PR #307, opened 2024-05-02
Make elan read from redis (and a bunch of other improvements to redis).

fische:
This PR makes elan read from Redis before hitting the bucket, so we can save on bucket operations. It does not make it write to redis as we were worried it would highly increase the number of evicted keys. We might experiment with that in the future.

There are a bunch of  additional improvements to the redis config as well:
* to avoid maintaining 2 separate sets of flags, I've moved the redis options from the mettle worker package to `//redis`;
* to avoid reimplementing the error rate limiter we use in the worker code, I implemented the Limiter interface from the redis package, so it is now supported natively within the clients;
* I've also made the redis MaxSize configurable, as we might want to tweak it at some point to make sure we use Redis to its full potential.

Those additional changes came in as I faced the issues they are trying to address. They are completely separate from the elan change and could be moved to a separate PR. If you think this PR is too big, I'm happy to split it.


mettle/worker/redis.go
```diff
@@ -38,17 +36,19 @@ var redisLatency = prometheus.NewHistogramVec(prometheus.HistogramOpts{
...
 // If readRedis is set, all reads will happen on this client. If not, everything
 // will go to the primary client.
-func newRedisClient(client elan.Client, primaryRedis, readRedis *redis.Client) elan.Client {
+func newRedisClient(client elan.Client, primaryRedis, readRedis *redis.Client, maxSize int64) elan.Client {
 	// This is a safeguard in case the caller does not pass readRedis.
 	if readRedis == nil {
 		readRedis = primaryRedis
 	}
+	if maxSize <= 0 {
+		maxSize = 200 * 1012 // 200 Kelly-Bootle standard units
```
Hamishpk, line 45: We have this set in two places but we will always want it to be the same (probably). Should we make this a exportable const in the redis library so we have maxSize = redis.DefaultMaxSize

fische: Ah yeah, good idea

Hamishpk (approved):

Hamishpk (approved):

PR #309, opened 2024-05-07
Dont take into account context canceled and rate limit errors in redis limiter.

fische:
I realised the redis limiter takes into account context cancellation errors and bumps the rate limiter if it sees that error. With mettle workers, this wasn't a big issue, but now elan hits redis as well, clients may cancel their requests.
It seems if `Allow` returns an error, the client will report it to the Limiter as well using `ReportResult` which I wasn't expecting, so I've added the rate limit error to the ones we exclude in `ReportResult`.
This PR also includes some tests for the limiter to make sure it actually works :D


ChangeLog
```diff
@@ -1,3 +1,10 @@
+Version 11.10.2
+--------------
+    * Add some tests to redis limiter.
+    * Dont take into account context canceled in limiter.
```
Liu-ko, line 4:
```suggestion
    * Don't take into account context canceled in limiter.
```

Liu-ko (approved):

Liu-ko (approved):

PR #311, opened 2024-05-16
Check caches in blobExist in the correct order.

fische:
I've just realised blobExists currently checks the existence of the blob in redis before checking it in the in-memory cache. Our redis currently gets hundreds of millions of EXISTS commands everyday, causing high CPU usage and spikes in latency. This will hopefully help with it and should help with the latency of Elan in general.


peterebden (approved):
I'm a little bit rusty on this but makes sense!

PR #313, opened 2024-06-15
Send back execution metadata with messages

peterebden:
Newer versions of REAPI have a field for this. We have the info so it's simple to attach.


fische (approved):
LGTM, thanks!
QQ: do we set the queued_timestamp somewhere? I can't see it in the API code, so I don't understand how the browser is able to say when the action was initially enqueued

peterebden:
I don't see it either, but I do see it showing up... not  sure tbh!

PR #317, opened 2024-07-11
Add gRPC API to Lucidity to list workers

peterebden:
This makes it possible to retrieve them programmatically. It also has Mettle workers detect & forward the hostname to Lucidity so it can be filtered by if there's one you care about particularly.


proto/lucidity/lucidity.proto
```diff
@@ -8,12 +8,16 @@ option go_package = "github.com/thought-machine/please-servers/proto/lucidity";
...
 service Lucidity {
   // Update sends the server an update on the current state of a worker.
-  rpc Update(UpdateRequest) returns (UpdateResponse);
+  rpc Update(Worker) returns (UpdateResponse);
+  // List the current set of workers
+  rpc ListWorkers(ListWorkersRequest) returns (ListWorkersResponse);
 }
 
-message UpdateRequest {
+message Worker {
```
peterebden, line 16: This adds quite a bit of noise to the PR but it seemed much nicer to name it this way, especially now there > 1 RPC in this file

fische (approved):

PR #321, opened 2024-09-23
Add some debug logs in flair.

fische:
This PR adds some debug logs to flair, so it's easier for us to see which client the load is coming from.

### DIFF
--- a/flair/rpc/rpc.go
+++ b/flair/rpc/rpc.go
@@ -136,6 +136,7 @@ func (s *server) GetCapabilities(ctx context.Context, req *pb.GetCapabilitiesReq
 }
 
 func (s *server) GetActionResult(ctx context.Context, req *pb.GetActionResultRequest) (ar *pb.ActionResult, err error) {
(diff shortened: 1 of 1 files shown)


peterebden:
I'm not sure how this will identify the client? In our usual setup the client of Flair is running on the same machine as it.
I also think it'll be very noisy when those clients are active...

fische:
> I'm not sure how this will identify the client? In our usual setup the client of Flair is running on the same machine as it. I also think it'll be very noisy when those clients are active...

Yeah actually you're right it's going to be noisy. We probably don't need the logs on reads, only on Execute; this should lower the noise. Then we can identify the machine through the labels on Kibana.

fische:
Actually I'm wondering if we could collect those as metrics instead of logs. That would be much more useful and wouldn't pollute the logs.

Tatskaari:
Let's have a think about this. What we want to know is: 
1) What machine is generated what actions for which targets
2) What actions are downloading the most data

I think we may want to make flair use structured logging so we can record metadata against these loglines that we can analyse in logstash too.

PR #325, opened 2024-11-24
Add a pre-flight action to ensure workers are able to execute actions

peterebden:
This seems useful while we're dealing with the Ubuntu Noble upgrade & machines that potentially aren't able to run things properly; this will cause them to die before they can become active & start poisoning the worker fleet.


mettle/main.go
```diff
@@ -76,8 +78,9 @@ var opts = struct {
 	} `command:"worker" description:"Start as a worker"`
 	Dual struct {
 		GRPC            grpcutil.Opts           `group:"Options controlling the gRPC server"`
-		Dir             string                  `short:"d" long:"dir" default:"plz-out/mettle" description:"Directory to run actions in"`
+		Dir             string                  `short:"d" long:"dir" default:"plz-out/mettle" description:"Base directory to run actions in"`
```
peterebden, line 81: Minor one I found while testing this: all the workers share the same temp directory in this setup. They don't normally clash because working dirs are named after the action digest, and we deduplicate actions, but it happens with the preflight check because it's the same one on each worker.

I don't think this would be an issue often otherwise, but I'm giving each its own directory for good measure.

mettle/worker/preflight.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/bazelbuild/remote-apis-sdks/go/pkg/uploadinfo"
+	pb "github.com/bazelbuild/remote-apis/build/bazel/remote/execution/v2"
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/known/durationpb"
+)
+
+// runPreflightAction runs a simple known action to ensure we can process them correctly.
+func (w *worker) runPreflightAction() error {
+	log.Notice("Preparing pre-flight action...")
+	const fileContents = "thirty-five ham and cheese sandwiches\n"
```
Hamishpk, line 17: This seems like a lot of sandwiches! It's probably fine but worth keeping an eye on it whilst this gets rolled out.

Hamishpk (approved):

mettle/worker/preflight.go
peterebden: Yup agreed. We should be able to roll  this out to just our staging environment first so it'll only be a few that get it.

Tatskaari: Right but if we're only rolling it out to staging, won't there be fewer consumers? Perhaps we should take this straight to prod to avoid wasted sandwiches.

PR #329, opened 2025-03-12
bypass limiter for action cache

mwalsh-gf:
Action cache look ups should  be much smaller than CAS read/writes so bypass the limiter for these operations.

These changes have come from the investigation into the slow elan cache - https://pennyworth.atlassian.net/browse/INFRA-98010

While bucket read times were low, the overall bucket latency was high, suggesting that requests were queuing to access the bucket. Increasing the parallelism improved latency, suggesting this was a bottleneck.

This aims to address the issue more efficiently by avoiding quick cheap operations queuing up behind slow expensive operations.


toastwaffle:
Please could you provide some context of why you want to make this change?

toastwaffle:
Test comment, please ignore

Hamishpk (approved):

mwalsh-gf:
hi

PR #340, opened 2026-03-06
Investigate Mettle timeout on premerge and postmerge tests

olha-ctrl:
Added logProcChain() to print child processes that keep worker in a waiting state: 

Logs from kibana indicates that  some pipes are still `open pipe:[850045] 2 -> pipe:[850046]`, so want to find out what they are. 

`
time="18:57:33.007" level=debug msg="Deadline exceeded: /proc snapshot" cmdline="/usr/local/bin/tm_sandbox bash --noprofile --norc -u -o pipefail -c export TMP_DIR=\"`pwd`\" TEST_DIR=\"`pwd`\" && export TEST=\"$TEST_DIR/notifications_service_sat\" && $TEST -kafka_provider=redpanda -db_provider=psql 2>&1 | tee $TMP_DIR/test.results && $TOOLS_GO tool covdata textfmt -i=. -o=$COVERAGE_FILE" fds="[0 -> /dev/null 1 -> pipe:[850045] 2 -> pipe:[850046]]" hash=9bbb0b015a5b9fcdbb53256e49444e65408bdac19921c68995c8f0d8a6d48314 pid=24877 wchan=do_wait
`


mettle/worker/worker.go
```diff
@@ -1076,6 +1077,102 @@ func (w *worker) logProcOnTimeout(pid int) {
...
+	firstChild := func(p int) int {
+		b, err := os.ReadFile(filepath.Join("/proc", strconv.Itoa(p), "task", strconv.Itoa(p), "children"))
+		if err != nil || len(b) == 0 {
+			return 0
+		}
+		f := strings.Fields(string(b))
+		if len(f) == 0 {
+			return 0
+		}
+		n, _ := strconv.Atoi(f[0])
```
scyyx5:
```suggestion
		n, err := strconv.Atoi(f[0])
        if err != nil {
            return 0
        }
```

mettle/worker/worker.go
```diff
@@ -1076,6 +1077,102 @@ func (w *worker) logProcOnTimeout(pid int) {
...
+				if f := strings.Fields(strings.TrimSpace(v)); len(f) > 0 {
+					return f[0]
+				}
+			}
+		}
+		return ""
+	}
+
+	cur := pid
+	parts := []string{fmt.Sprintf("%d:%s", cur, name(cur))}
```
scyyx5: Nit: Since we already know the max length, consider preallocating `parts`     e.g. make([]string, 0, maxDepth+1).

mettle/worker/worker.go
```diff
@@ -1076,6 +1077,102 @@ func (w *worker) logProcOnTimeout(pid int) {
...
+	readTrim := func(path string) string {
+		b, err := os.ReadFile(path)
+		if err != nil || len(b) == 0 {
+			return ""
+		}
+		return strings.TrimSpace(string(b))
+	}
+
+	name := func(p int) string {
+		n := readTrim(filepath.Join("/proc", strconv.Itoa(p), "comm"))
```
scyyx5: nit: avoid repeating `filepath.Join("/proc", strconv.Itoa(p)` multiple times. 
```suggestion
        pidStr := strconv.Itoa(p)
		n := readTrim(filepath.Join("/proc", pidStr, ...))
```

mettle/worker/worker.go
olha-ctrl: make sense, updated!

mettle/worker/worker.go
olha-ctrl: Updated

mettle/worker/worker.go
olha-ctrl: Updated

mettle/worker/worker.go
```diff
@@ -1076,6 +1077,109 @@ func (w *worker) logProcOnTimeout(pid int) {
 	}).Debug("Deadline exceeded: /proc snapshot")
 }
 
+// logProcChain logs a compact parent→child execution chain for the given PID
+// This helps quickly identify which subprocess is still running when the worker
+// times out
+func (w *worker) logProcChain(pid int) {
+	const maxDepth = 10
+
+	readTrim := func(path string) string {
```
toastwaffle: I don't like these functions-as-variables - is there a specific reason you're using them? if not, please just make them normal top-level functions (and consider adding doc comments)

mettle/worker/worker.go
```diff
@@ -1076,6 +1077,109 @@ func (w *worker) logProcOnTimeout(pid int) {
...
+		}
+		return ""
+	}
+
+	cur := pid
+	parts := make([]string, 0, maxDepth+1)
+	parts = append(parts, fmt.Sprintf("%d:%s", cur, name(cur)))
+
+	for i := 0; i < maxDepth; i++ {
+		c := firstChild(cur)
```
toastwaffle: what if the problematic child is not the first child?

mettle/worker/worker.go
```diff
@@ -1076,6 +1077,109 @@ func (w *worker) logProcOnTimeout(pid int) {
...
+		"tail_name":          name(tailPID),
+		"tail_state":         tailState,
+		"tail_state_comment": stateComment(tailState),
+		"tail_wchan":         readTrim(filepath.Join("/proc", strconv.Itoa(tailPID), "wchan")),
+	}).Debug("Timeout: process chain")
+}
+
+// stateComment is a helper to translate tail_state code into text
+func stateComment(s string) string {
+	switch s {
```
toastwaffle, line 1216: Missing `I (idle)`

mettle/worker/worker.go
```diff
@@ -1076,6 +1077,109 @@ func (w *worker) logProcOnTimeout(pid int) {
...
+	logr.WithFields(logrus.Fields{
+		"hash":               w.actionDigest.Hash,
+		"pid":                pid,
+		"chain":              strings.Join(parts, " -> "),
+		"tail_pid":           tailPID,
+		"tail_name":          name(tailPID),
+		"tail_state":         tailState,
+		"tail_state_comment": stateComment(tailState),
+		"tail_wchan":         readTrim(filepath.Join("/proc", strconv.Itoa(tailPID), "wchan")),
+	}).Debug("Timeout: process chain")
```
toastwaffle:
```suggestion
	}).Debug("Deadline exceeded: process chain")
```

For consistency with `logProcOnTimeout`

mettle/worker/worker.go
```diff
@@ -1076,6 +1077,109 @@ func (w *worker) logProcOnTimeout(pid int) {
...
+	}
+
+	cur := pid
+	parts := make([]string, 0, maxDepth+1)
+	parts = append(parts, fmt.Sprintf("%d:%s", cur, name(cur)))
+
+	for i := 0; i < maxDepth; i++ {
+		c := firstChild(cur)
+		if c == 0 {
+			break
```
toastwaffle: We should add an explicit marker for whether we hit this case or not

toastwaffle (changes requested):

mettle/worker/worker.go
olha-ctrl: Moved functions-as-variables to proper functions; I wanted to make a function self-contained, do it's easy to delete it after we completed investigations. But I agree it doesn't look good.

mettle/worker/worker.go
olha-ctrl: good point, I added loop for all child processes and shortened the log message to keep kibana logs clean

mettle/worker/worker.go
olha-ctrl: Added

mettle/worker/worker.go
olha-ctrl: added
```
	if childrenSummary != "" {
		fields["children"] = childrenSummary
	} else {
		fields["children"] = "none"
	}
```

mettle/worker/worker.go
olha-ctrl: Updated!

Hamishpk (approved):

mettle/worker/worker.go
```diff
@@ -1076,6 +1077,147 @@ func (w *worker) logProcOnTimeout(pid int) {
...
+					fmt.Sprintf("%d:%s:%s(%s)",
+						c,
+						procName(c),
+						state,
+						stateComment(state),
+					),
+				)
+			}
+			childrenSummary = strings.Join(items, " | ")
+			break
```
toastwaffle: Doesn't this mean we now stop exploring deeply if we find multiple children? that also seems impractical; shouldn't we be outputting the whole tree?

olha-ctrl: I was trying to log minimal useful info to keep kibana logs is a proper state, but you are right. I updated function to log full process-tree in the compact format:
```
24877:bash[S - sleeping/waiting]{do_wait} -> 24891:tee[S - sleeping/waiting]{pipe_read}
```

mettle/worker/worker.go
```diff
@@ -1067,15 +1076,158 @@ func (w *worker) logProcOnTimeout(pid int) {
...
+		for _, c := range children {
+			if nodes >= maxNodes {
+				return
+			}
+			walk(c, depth+1, path)
+		}
+	}
+
+	startDepth := 0
+	walk(pid, startDepth, nil)
```
toastwaffle:
```suggestion
	walk(pid, 0, nil)
```

Nit: no need for the single use variable

mettle/worker/worker.go
```diff
@@ -1067,15 +1076,158 @@ func (w *worker) logProcOnTimeout(pid int) {
...
+//
+// 24877:bash[S - sleeping/waiting]{do_wait} -> 24890:test_binary[D - io_wait)]{tcp_recvmsg}
+// 24877:bash[S - sleeping/waiting]{do_wait} -> 24891:tee[S - sleeping/waitin]{pipe_read}
+func (w *worker) logProcChain(pid int) {
+	const (
+		maxDepth = 6
+		maxNodes = 40
+	)
+	lines := make([]string, 0, maxNodes)
+	nodes := 0
```
toastwaffle, line 1109: nit: you could skip this variable, and just use len(lines)

mettle/worker/worker.go
```diff
@@ -1036,21 +1037,29 @@ func (w *worker) logProcOnTimeout(pid int) {
 	const maxTargetLen = 256
 	p := strconv.Itoa(pid)
 	base := filepath.Join("/proc", p)
+	var errs []string
 
 	cmdlineBytes, err := os.ReadFile(filepath.Join(base, "cmdline"))
 	cmdline := ""
-	if err == nil && len(cmdlineBytes) > 0 {
+	if err != nil {
+		errs = append(errs, "cmdline")
```
toastwaffle:
```suggestion
		errs = append(errs, fmt.Errorf("cmdline: %v", err))
```

similarly below

toastwaffle (approved):

mettle/worker/worker.go
olha-ctrl: fixed

mettle/worker/worker.go
olha-ctrl: updated to use `var errs []error` instead of `var errs []string` as was suggested

PR #351, opened 2026-04-23
Address vulnerabilities in Mettle repo

scyyx5:
1. Cherry-pick dependency-update commits to bump dependencies to latest version and fix merge conflict. 
2. Switches health registration to the official grpc health server implementation compatible with newer grpc
3. Adds the missing grpc health build dependency
4. Fixes a go.sum module path typo for cloud.google.com/go/longrunning
Note: In grpc v1.79.3, the `HealthServer` interface added a new List method and changed the Watch signature, making existing custom health server implementations incompatible. So this PR replace direct registration of custom servers with the official `health.NewServer()` implementation and update `BUILD` dependencies accordingly.


scyyx5:
Skip [Bump golang.org/x/net from 0.19.0 to 0.38.0](https://github.com/thought-machine/please-servers/pull/349) as it already bumped to v0.47.0

flair/rpc/rpc.go
```diff
@@ -73,23 +74,6 @@ type server struct {
 	timeout                                    time.Duration
 }
 
-func (s *server) Check(context.Context, *hpb.HealthCheckRequest) (*hpb.HealthCheckResponse, error) {
```
toastwaffle, line 76: Did Hamish suggest you could remove this? I don't think it's safe to just switch to health.NewServer() - it does not have the same behaviour as this Check function.

What you should do is embed `hpb.UnimplementedHealthServer` into the `server` struct, so that we automatically get stub methods for the endpoints we don't intend to implement.

toastwaffle (changes requested):
This PR needs to have changes to `third_party/go/BUILD` in it - you should be able to use `plz puku sync -w`

flair/rpc/rpc.go
```diff
@@ -73,7 +78,12 @@ type server struct {
...
+type healthServer struct {
```
toastwaffle, line 84: Why are you adding this new struct instead of continuing to use the existing server struct? You should be fine to embed multiple `UnimplementedWhateverServer`s, so long as their method names don't conflict

If they do conflict, please add a comment to explain that

flair/rpc/rpc.go
```diff
@@ -86,7 +96,7 @@ func (s *server) Check(context.Context, *hpb.HealthCheckRequest) (*hpb.HealthChe
 	return &hpb.HealthCheckResponse{Status: hpb.HealthCheckResponse_SERVING}, nil
 }
 
-func (s *server) Watch(*hpb.HealthCheckRequest, hpb.Health_WatchServer) error {
+func (s *healthServer) Watch(*hpb.HealthCheckRequest, hpb.Health_WatchServer) error {
```
toastwaffle: Watch can be removed, since `UnimplementedHealthServer` will provide an equivalent method

toastwaffle (changes requested):

third_party/go/BUILD
```diff
@@ -4,8 +4,8 @@ package(default_visibility = ["PUBLIC"])
 
 go_toolchain(
     name = "toolchain",
-    version = "1.23.2",
     install_std = False,
+    version = "1.24.0",
```
toastwaffle, line 8: One for a later PR, but we should get this to 1.26.2

flair/rpc/rpc.go
scyyx5: added a comment

toastwaffle (approved):

PR #352, opened 2026-04-24
Add platform properties functionality to mettle workers

scyyx5:


mettle/worker/worker.go
```diff
@@ -562,6 +567,43 @@ func (w *worker) runTaskRequest(req *pb.ExecuteRequest) *pb.ExecuteResponse {
...
+}
+
+func (w *worker) logPlatformProperties() {
+	if len(w.platformProps) == 0 {
+		return
+	}
+	logr.WithFields(logrus.Fields{
+		"hash":               w.actionDigest.Hash,
+		"platformProperties": formatPlatformProperties(w.platformProps),
+	}).Info("Action specifies platform properties; would use them to select sandbox filesystem")
```
Hamishpk, line 604: nit: platform properties may be used for many things so I would shorten this just to list the platform properties and not mention the sandbox

mettle/worker/worker_platform_test.go
```diff
@@ -0,0 +1,39 @@
...
+func TestCopyPlatformProperties(t *testing.T) {
+	original := map[string]string{
+		"filesystem":  "disk",
+		"worker_pool": "large",
+	}
+	copied := copyPlatformProperties(original)
+
+	assert.Equal(t, original, copied)
+
+	original["filesystem"] = "tmpfs"
```
Hamishpk, line 18: Do we need this line here?

Hamishpk (commented):
Two small nits but otherwise looks good to me

PR #354, opened 2026-06-01
Revert mettle temp debug  logs

olha-ctrl:
The original issue causing timeouts has been fixed. This change removes the temporary logging that was added to help investigate the hanging processes.


mettle/worker/worker.go
toastwaffle: plz fix lint

toastwaffle (commented):
Do we actually want to remove this? how much of a cost are we paying for this remaining here?

mettle/worker/worker.go
olha-ctrl: Fixed the linter. I had a chat with Hamish and we decided to remove it as it's ~300lines on code; if we need it one day, we can revert this change

Hamishpk: Can we remove this whole go routine as without the logging it will no longer do anything

olha-ctrl: done!

Hamishpk (approved):

PR #359, opened 2026-07-10
Upgrade the Go toolchain, packages and Please plugin

toastwaffle:
Mostly a `go get -u ./...` (after a `plz clean -f`), followed by `go mod tidy` and `plz puku sync -w`, and then fixing stuff.


elan/rpc/BUILD
```diff
@@ -60,7 +60,7 @@ go_test(
 genrule(
     name = "test_data",
     cmd = [
-        'for i in `seq 1 10000`; do echo "$i ----------------------------------------- $i" >> "tmp.txt"; done',
+        'for i in `seq 1 100000`; do echo "$i ----------------------------------------- $i" >> "tmp.txt"; done',
```
toastwaffle, line 63: Some new package caused the test server to return more bytes in a single `Recv` call, so I had to make the test file bigger

elan/rpc/rpc.go
```diff
@@ -630,6 +630,14 @@ func (s *server) QueryWriteStatus(ctx context.Context, req *bs.QueryWriteStatusR
 	return nil, status.Errorf(codes.NotFound, "write %s not found", req.ResourceName)
 }
 
+func (s *server) SpliceBlob(ctx context.Context, req *pb.SpliceBlobRequest) (*pb.SpliceBlobResponse, error) {
```
toastwaffle, line 633: New methods in the proto

Hamishpk (approved):
Thanks :)

PR #363, opened 2026-09-19
Update reported max API version to latest

peterebden:
Handle newer symlink field for a client that sends that, and populate the new `root_directory_digest` field for a client that wants it.

Not much to do otherwise, we already supported most of the features of earlier versions (like compression) before they got (retroactively) tagged. There are other changes, mostly around multiple digest functions and split-and-splice but they're opt-in and we're a SHA-256 only server so we don't have to support any of them.

Update to a more recent Go version while I'm here.

I have not upgraded `remote-apis-sdks`. There are various changes in it which are IMO highly questionable (silently ignoring unreadable outputs and aggressively opening 25 parallel connections to the CAS server) and relatively little that looks particularly useful for us.


mettle/api/api.go
```diff
@@ -291,7 +291,7 @@ func (s *server) GetCapabilities(ctx context.Context, req *pb.GetCapabilitiesReq
 			ExecEnabled:    true,
 		},
 		LowApiVersion:  &semver.SemVer{Major: 2, Minor: 0},
-		HighApiVersion: &semver.SemVer{Major: 2, Minor: 1}, // optimistic
```
peterebden, line 294: our optimism here eventually paid off and it did get tagged, quite some time after this comment.

AGENTS.md
```diff
@@ -0,0 +1,254 @@
+# AGENTS.md
```
peterebden, line 1: Adding this while I'm here

mettle/worker/outputs_test.go
```diff
@@ -0,0 +1,87 @@
...
+	outDir := ar.OutputDirectories[0]
+	require.NotNil(t, outDir.RootDirectoryDigest)
+
+	// The root digest should be that of the Tree's root, which carries the pack property.
+	tree := &pb.Tree{}
+	require.NoError(t, proto.Unmarshal(blob(t, entries, outDir.TreeDigest), tree))
+	rootDigest, err := digest.NewFromMessage(tree.Root)
+	require.NoError(t, err)
+	assert.Equal(t, rootDigest.ToProto(), outDir.RootDirectoryDigest)
+	assert.NotEqual(t, digest.Digest{}, rexclient.PackDigest(tree.Root))
```
toastwaffle, line 39: Why do we only assert that it is non-zero, but not assert what it actually is?

mettle/worker/outputs_test.go
```diff
@@ -0,0 +1,87 @@
...
+	root := &pb.Directory{}
+	require.NoError(t, proto.Unmarshal(blob(t, entries, outDir.RootDirectoryDigest), root))
+	require.Len(t, root.Files, 1)
+	assert.Equal(t, "a.txt", root.Files[0].Name)
+	require.Len(t, root.Directories, 1)
+	assert.Equal(t, "sub", root.Directories[0].Name)
+	sub := &pb.Directory{}
+	require.NoError(t, proto.Unmarshal(blob(t, entries, root.Directories[0].Digest), sub))
+	require.Len(t, sub.Files, 1)
+	assert.Equal(t, "b.txt", sub.Files[0].Name)
```
toastwaffle, line 51: `require.Len(t, sub.Directories, 0)`? Or not worth it?

toastwaffle: or maybe `assert.Empty`

mettle/worker/outputs_test.go
```diff
@@ -0,0 +1,87 @@
...
+	// No symlinks must stay nil so the serialised ActionResult doesn't change.
+	assert.Nil(t, allOutputSymlinks(&pb.ActionResult{}))
+
+	// Only the deprecated fields are populated
+	assert.Equal(t, []*pb.OutputSymlink{file, dir}, allOutputSymlinks(&pb.ActionResult{
+		OutputFileSymlinks:      []*pb.OutputSymlink{file},
+		OutputDirectorySymlinks: []*pb.OutputSymlink{dir},
+	}))
+
+	// output_symlinks is already populated, which takes priority
```
toastwaffle, line 81: This test doesn't actually assert that `output_symlinks` takes priority; `allOutputSymlinks` could still read the deprecated fields and pass this test.

AGENTS.md
toastwaffle: sneaky sneaky!

toastwaffle (approved):

